PR #2, opened 2025-01-24
Add configs and stuff

lewtun:


StEvUgnIn:
Shouldn't you add DeepSpeed as a dependency?

PR #33, opened 2025-01-25
Add devcontainer configuration for VS Code

bhack:


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM pytorch/pytorch:2.5.1-cuda12.4-cudnn9-runtime
+
+# Set the working directory
```
pepinu, line 3: LLM generated comments when the Dockerfile is readable add bloat

requirements.txt
```diff
@@ -0,0 +1,24 @@
...
+liger_kernel==0.5.2
+lighteval @ git+https://github.com/huggingface/lighteval.git@4f381b352c0e467b5870a97d41cb66b487a2c503#egg=lighteval[math]
+math-verify
+packaging>=23.0
+parameterized>=0.9.0
+pytest
+safetensors>=0.3.3
+sentencepiece>=0.1.99
+torch>=2.5.1
+transformers @ git+https://github.com/huggingface/transformers.git@main
```
pepinu, line 21: Why compiling it from repo?

requirements.txt
```diff
@@ -0,0 +1,24 @@
...
+trl @ git+https://github.com/huggingface/trl.git@main
```
pepinu, line 22: same as above

pepinu (commented):
a few comments otherwise looks alright

Dockerfile
bhack: https://github.com/huggingface/open-r1/blob/main/setup.py#L64

requirements.txt
bhack: https://github.com/huggingface/open-r1/blob/main/setup.py#L64

requirements.txt
bhack: https://github.com/huggingface/open-r1/blob/main/setup.py#L65

Dockerfile
bhack: Yes it was added by GitHub workspace in the PR check but we could remove it now

requirements.txt
lewtun, line 1: Maintaining two sets of requirements (`setup.py` and this file) will be a bit annoying to maintain. Is it possible to set this up by running `pip install -e '.[dev]'`?

bhack: Do you prefer this or to autogenerate the file with [`pip-compile`](https://pypi.org/project/pip-tools/)?

bhack: I've updated it but just a more general reminder, with you preferred solution on modern `pip` we have
```python
 DEPRECATION: Legacy editable install of open-r1[dev]==0.1.0.dev0 from file:///workspaces/open-r1 (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457
 ```

File: Dockerfile [hunk 4]
```diff
@@ -0,0 +1,4 @@
+FROM pytorch/pytorch:2.5.1-cuda12.4-cudnn9-runtime
```
bhack, line 1: @lewtun Do we need to switch to `12.1` for https://github.com/huggingface/open-r1/pull/121 ?

ctjlewis: I'd think so, did you manage to actually reproduce anything on the 14.1? I looked at it and really didn't see how it would actually run, when I ran on 12.4 it gave a segmentation fault.

PR #52, opened 2025-01-26
Add Environment Test Script

sambhavnoobcoder:
During my initial setup, I found myself writing a quick test to verify if my environment was properly configured on my local . This simple script checks PyTorch installation, model loading, and basic inference - basically making sure everything's working before diving into actual development. Sharing it in case others find it useful, but feel free to discard the PR if it doesn't fit the project's needs . 

also kudos on the initiative . this is really awesome !!

### DIFF
--- a/test_installation.py
+++ b/test_installation.py
@@ -1,0 +1,75 @@
+from transformers import AutoModelForCausalLM, AutoTokenizer
+import torch
+import time
+from tqdm import tqdm
+import os
+
+os.environ["TOKENIZERS_PARALLELISM"] = "false"
+
+def test_installation():
+    print("Testing installation...")
+    start_time = time.time()
+    
+    # Test PyTorch
+    print("\n1. Testing PyTorch:")
+    print(f"PyTorch version: {torch.__version__}")
+    print(f"CUDA available: {torch.cuda.is_available()}")
+    print(f"MPS available: {torch.backends.mps.is_available()}")
+    
+    # Test Transformers
+    print("\n2. Testing Transformers:")
+    model_name = "Qwen/Qwen2.5-Math-1.5B-Instruct"
+    print(f"Loading model and tokenizer from {model_name}")
+    
+    try:
+        print("\nLoading tokenizer...")
+        tokenizer = AutoTokenizer.from_pretrained(model_name)
+        print("✓ Tokenizer loaded successfully")
+        
+        print("\nLoading model (this may take several minutes on first run)...")
+        model = AutoModelForCausalLM.from_pretrained(
+            model_name,
+            torch_dtype=torch.float16,
+            device_map="auto"
+        )
+        print("✓ Model loaded successfully")
+        
+        # Test inference
+        prompt = "What is 2 + 2? Please reason step by step, and put your final answer within \\boxed{}."
+        print("\n3. Testing inference:")
+        print(f"Prompt: {prompt}")
+        
+        print("\nGenerating response (this may take a minute on first run)...")
+        inference_start = time.time()
+        
+        inputs = tokenizer(prompt, return_tensors="pt").to(model.device)
+        print(f"Input processing complete. Device: {model.device}")
+        
+        # Reduced complexity generation
+        max_tokens = 50  # Reduced from 100
+        print(f"\nGenerating up to {max_tokens} tokens with simplified settings...")
+        outputs = model.generate(
+            **inputs,
+            max_new_tokens=max_tokens,
+            do_sample=False,  # Deterministic generation
+            temperature=0.0,  # Remove randomness
+            num_return_sequences=1,
+            pad_token_id=tokenizer.eos_token_id,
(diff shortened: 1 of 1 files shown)


qgallouedec:
Thanks for the suggestion. How is it specific to Open-R1? It seems more related to transformers.
Usually we tests using unittest and a CI, I'd prefer this option

sambhavnoobcoder:
Thanks for the feedback! I completely agree with your points about unittest and CI being the proper approach.

However, I had a slightly different use case in mind for this script. While we work towards building a robust test suite, this simple verification tool could help new contributors quickly validate their environment setup when they first clone the repo. I've found it particularly helpful for catching common setup issues before diving into development.

That said, I see this as a temporary solution. Once we have comprehensive unittest coverage and CI in place, we can deprecate this script in favor of proper testing infrastructure.

Would love to hear your thoughts on this perspective. Happy to take a different approach if you feel the focus should be on building out the test suite first.

test_installation.py
```diff
@@ -0,0 +1,75 @@
...
+            num_return_sequences=1,
+            pad_token_id=tokenizer.eos_token_id,
+            eos_token_id=tokenizer.eos_token_id
+        )
+        
+        inference_time = time.time() - inference_start
+        response = tokenizer.decode(outputs[0], skip_special_tokens=True)
+        print(f"\nModel response:\n{response}")
+        
+        total_time = time.time() - start_time
```
pepinu, line 65: Are you 100% sure this does not need cuda synchronize clocks [1]?
I am not sure here since there is a decoding step which takes the response and prints it so it is probably okay, just an off the cuff thought that I had.

[1]: https://www.speechmatics.com/company/articles-and-news/timing-operations-in-pytorch

test_installation.py
```diff
@@ -0,0 +1,75 @@
+from transformers import AutoModelForCausalLM, AutoTokenizer
+import torch
+import time
+from tqdm import tqdm
+import os
+
+os.environ["TOKENIZERS_PARALLELISM"] = "false"
+
+def test_installation():
```
pepinu, line 9: This is nice, however would be great if we could parametrize this for better maintainability.  Like, model_name, device, prompt, verbose flag (for detailing time info) could be passed in the function definition.

This way it will be easier to separate concerns when MPS does not work for some scenario, CUDA fails for some version, etc.

test_installation.py
```diff
@@ -0,0 +1,75 @@
...
+        
+        print("\nLoading model (this may take several minutes on first run)...")
+        model = AutoModelForCausalLM.from_pretrained(
+            model_name,
+            torch_dtype=torch.float16,
+            device_map="auto"
+        )
+        print("✓ Model loaded successfully")
+        
+        # Test inference
```
pepinu, line 37: you have the one liner comments provided presumably by one of LLMs, they are a great hint which sections could be abstracted to test methods that could be used in the future for different installation configurations. Like test_inference could be taken out of this function so we have test setup - loading model and setting up the prompt, running inference, displaying the result.

test_installation.py
sambhavnoobcoder: yes , i agree . I've parametrized the function to make it more flexible and maintainable in commit [eeb3a1d](https://github.com/huggingface/open-r1/pull/52/commits/eeb3a1d6c9235afcbf16e1172abe59e2bc0f7424)

test_installation.py
sambhavnoobcoder: Okay ,now I've refactored the code to separate concerns into distinct test methods:

1. `_test_pytorch_environment()`: [2c486a8](https://github.com/huggingface/open-r1/pull/52/commits/2c486a88e9a0ae50bd2979d0693abfd15e23887c) commit 
2. `_load_model_and_tokenizer()`: [fef029b](https://github.com/huggingface/open-r1/pull/52/commits/fef029b2335ae9779a0209ebec4c3138c2c9963c) commit
3. `_run_inference()`: [20d4f08](https://github.com/huggingface/open-r1/pull/52/commits/20d4f084673c8d6e72638ff8911c6e77a2579528) commit

test_installation.py
sambhavnoobcoder: yes , even though here `tokenizer.decode()` does serve as an implicit synchronisation of sorts , but it is better to make a robust and explicit behaviour rather than relying on implicit behaviour , hence this is  addressed in commit [005fc7c](https://github.com/huggingface/open-r1/pull/52/commits/005fc7ccb5da0349a2fe547f9410dc588a90f9a9)

sambhavnoobcoder:
resolved all comments , kindly have a look and inform me of any more required changes .

qgallouedec:
I still don't quite understand how this is specific to Open-R1. It loads a model and runs a generation if I understand correctly. So it's a demo code for the transformers library basically? Maybe it would make more sense as a transformers lib demo script?
I think this script can stay in this branch, so we can refer to it if needed, but I wouldn't merge it in main to keep the repo as simple as possible.

Issue #86, opened 2025-01-28
Issue: Potential Inflation of Reward Metrics for Unparseable Gold Solutions


agulati18:
**Description**

In the def_accuracy(reward) function, the current implementation assigns a reward of 1.0 when the gold solution cannot be parsed. This artificially inflates the reported completion metric, as it effectively rewards the model regardless of performance in those cases.

This issue is observed in the following code snippet:

[Link to Code](https://github.com/huggingface/open-r1/blob/e2e403daffdfc4c9f62e10af609faa491866aac7/src/open_r1/grpo.py#L41)

```
if gold_parsed is not None:
    reward = float(verify(answer_parsed, gold_parsed))
else:
    reward = 1.0  # Artificially inflates metrics
    print("Failed to parse gold solution: ", sol)
rewards.append(reward)
```

**Impact**
Resolving this issue will:

1. Provide a more accurate representation of model performance
2. Improve transparency by differentiating between successful completions and unparseable examples
3. Allow better debugging of systemic issues with the dataset or parsing logic

**Problems with the Current Implementation**

1. Metric Inflation: Assigning 1.0 for unparseable gold solutions treats these cases as if the model performed perfectly, leading to an inaccurate eval
2. Masking Failures: Legitimate issues with the dataset, parsing logic, or model predictions may go unnoticed
3. Non-Informative Metrics: The current approach fails to differentiate between examples with valid gold solutions and those without

**Proposed Solutions**

1. Skip Unparseable Examples: Exclude examples with unparseable gold solutions from both the numerator and denominator of the accuracy metric.

```
if gold_parsed is not None:
    reward = float(verify(answer_parsed, gold_parsed))
else:
    print("Failed to parse gold solution: ", sol)
    continue  # Skip this example entirely
rewards.append(reward)
```

2. Neutral or Penalised Reward: Assign a neutral or penalised reward (e.g., 0.5 or 0.0) for unparseable gold solutions instead of 1.0.

```
if gold_parsed is not None:
    reward = float(verify(answer_parsed, gold_parsed))
else:
    reward = 0.5  # Neutral reward
    print("Failed to parse gold solution: ", sol)
rewards.append(reward)
```

3. Flag and Investigate Failures: Log all instances of unparseable gold solutions for further analysis. This would help identify whether the issue lies with the dataset, parsing logic, or other systemic problems.

```
if gold_parsed is not None:
    reward = float(verify(answer_parsed, gold_parsed))
else:
[...]

PR #87, opened 2025-01-28
Solution for Potential Inflation of Reward Metrics for Unparseable Go…

agulati18:
[Solution Pathway #2 in Issue #86](https://github.com/huggingface/open-r1/issues/86)

### **Recap:**
In the def_accuracy(reward) function, the current implementation assigns a reward of 1.0 when the gold solution cannot be parsed. This approach artificially inflates the reported completion metric, as it rewards the model regardless of its actual performance.

[Link to Code](https://github.com/huggingface/open-r1/blob/e2e403daffdfc4c9f62e10af609faa491866aac7/src/open_r1/grpo.py#L41) 

```
if gold_parsed is not None:
    reward = float(verify(answer_parsed, gold_parsed))
else:
    reward = 1.0  # Artificially inflates metrics
    print("Failed to parse gold solution: ", sol)
rewards.append(reward)
```

### **Proposed Fix in This PR:**
Assign a neutral reward of 0.5 for unparseable gold solutions instead of 1.0. This adjustment provides a more balanced evaluation of model performance while still acknowledging the ambiguity of such cases.

```
if gold_parsed is not None:
    reward = float(verify(answer_parsed, gold_parsed))
else:
    reward = 0.5  # Neutral reward
    print("Failed to parse gold solution: ", sol)
rewards.append(reward)
```

### DIFF
--- a/src/open_r1/grpo.py
+++ b/src/open_r1/grpo.py
@@ -68,8 +68,8 @@ def accuracy_reward(completions, solution, **kwargs):
             # Reward 1 if the content is the same as the ground truth, 0 otherwise
             reward = float(verify(answer_parsed, gold_parsed))
         else:
-            # If the gold solution is not parseable, we reward 1 to skip this example
-            reward = 1.0
+            # If the gold solution is not parseable, we reward a neutral reward (0.5) to skip this example
+            reward = 0.5  # Neutral reward
             print("Failed to parse gold solution: ", sol)
         rewards.append(reward)
 


mradityagoyal:
Can you elaborate on why this should be a neutral 0.5 and not something lower?

agulati18:
The decision to allocate a score of 0.5 for unparseable optimal answers was intended to have a neutral impact, ensuring the model is neither penalised nor rewarded for examples we cannot evaluate.

Alternatively, we could exclude such examples entirely, omitting them from the evaluation and averaging performance only over the subset of cases where cross-referencing is possible.

mradityagoyal:
Thanks. 
I am not sure if 0.5 is neutral either. My vote is to skip those examples altogether... But I am not sure how to do that.

agulati18:
Here’s how we might skip unparseable examples, as discussed in [the issue I raised](https://github.com/huggingface/open-r1/issues/86):

```
if gold_parsed is not None:
    reward = float(verify(answer_parsed, gold_parsed))
else:
    print("Failed to parse gold solution: ", sol)
    continue  # Skip this example entirely
rewards.append(reward)
```

Even so, I use a neutral reward (e.g. 0.5) instead of skipping unparseable gold solutions to ensure that evaluation remains fair, accurate, and diagnostically useful.

Simply skipping these cases introduces bias, distorts model accuracy, and makes comparisons between models unreliable.

When unparseable examples are removed from both the numerator and denominator, the accuracy metric no longer reflects the model's real-world performance. If these cases correlate with complexity, skipping them artificially inflates accuracy by removing difficult examples that the model might actually. systematically fail on. This creates a misleading picture of performance, hiding failure patterns rather than addressing them

Skipping also breaks comparability between models. If two models encounter different numbers of unparseable cases, their accuracy scores become calculated over different denominators, making direct comparison unreliable. A model with fewer parsing failures may appear stronger, even if its actual performance on valid cases is the same or worse. 

By using a neutral reward, we ensure that all models are evaluated consistently, preventing unfair comparisons.

PR #91, opened 2025-01-28
Change conda for uv

andimarafioti:
Adapted the virtual environment to use `uv` instead of `conda`. Some of the installs were bugging with `uv`, so I kept `pip` for those, but install speed is the same between using `pip` in `conda` or `pip` in `uv`. 
I tested this by following the readme in the cluster and launching the first accelerate command, which works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ uv pip install vllm==0.6.6.post1
 
 # For HF (cluster only has CUDA 12.1)
 pip install vllm==0.6.6.post1 --extra-index-url https://download.pytorch.org/whl/cu121
-export LD_LIBRARY_PATH=/path/to/open-r1/openr1/lib/python3.11/site-packages/nvidia/nvjitlink/lib:$LD_LIBRARY_PATH  # Change this path to the actual path on your system
+export LD_LIBRARY_PATH=$(python -c "import site; print(site.getsitepackages()[0] + '/nvidia/nvjitlink/lib')"):$LD_LIBRARY_PATH
 ```
 
 This will also install PyTorch `v2.5.1` and it is **very important** to use this version since the vLLM binaries are compiled for it. You can then install the remaining dependencies for your specific use case via `pip install -e .[LIST OF MODES]`. For most contributors, we recommend:
--- a/README.md
+++ b/README.md
@@ -31,16 +31,17 @@ We will use the DeepSeek-R1 [tech report](https://github.com/deepseek-ai/DeepSee
 To run the code in this project, first, create a Python virtual environment using e.g. Conda:
 
 ```shell
-conda create -n openr1 python=3.11 && conda activate openr1
+uv venv openr1 --python 3.11 && source openr1/bin/activate  && uv pip install pip
 ```
 
 Next, install vLLM:
 
 ```shell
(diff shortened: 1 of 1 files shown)


pepinu:
Whats the reason?

qgallouedec:
I agree with your decision to stick with `pip` in the README. While `uv` is indeed faster, `pip` is the default and standard package manager for Python, which ensures compatibility and minimizes extra steps for users. Not everyone may have `uv` installed or be familiar with it, so keeping `pip` makes the guide more universally accessible. Users who prefer `uv` likely already know how to switch.

andimarafioti:
But you're using conda in the README 😅 If you would prefer to use purely pip, I can change to create the virtual env with pip directly, would that be better?

qgallouedec:
Fair enough

qgallouedec (approved):

README.md
```diff
@@ -31,16 +31,17 @@ We will use the DeepSeek-R1 [tech report](https://github.com/deepseek-ai/DeepSee
 To run the code in this project, first, create a Python virtual environment using e.g. Conda:
```
lewtun, line 31:
```suggestion
To run the code in this project, first, create a Python virtual environment using e.g. [uv](https://github.com/astral-sh/uv):
```

README.md
```diff
@@ -31,16 +31,17 @@ We will use the DeepSeek-R1 [tech report](https://github.com/deepseek-ai/DeepSee
...
 
 Next, install vLLM:
 
 ```shell
-pip install vllm==0.6.6.post1
+uv pip install vllm==0.6.6.post1
 
 # For HF (cluster only has CUDA 12.1)
 pip install vllm==0.6.6.post1 --extra-index-url https://download.pytorch.org/whl/cu121
+export LD_LIBRARY_PATH=/path/to/open-r1/openr1/lib/python3.11/site-packages/nvidia/nvjitlink/lib:$LD_LIBRARY_PATH  # Change this path to the actual path on your system
```
lewtun, line 44: This is pretty clunky, is it something specific to the HF cluster or needed more generally?

lewtun (approved):
LGTM with a nit and question about the LD path

README.md
andimarafioti: I've seen this happen in a bunch of different setups, but I for sure have this issue in the HF cluster all the time. I thought everyone was doing this, I'll ask in the channel.

andimarafioti: yep, actually you are also having this issue while running this code in the cluster and Mathieu already suggested adding this as a solution 😅

qgallouedec: Maybe this to be more flexible ?
https://github.com/pytorch/pytorch/issues/111469#issuecomment-2497452760

lewtun: Chatted internally, seems specific to our cluster.

andimarafioti: I love quentin's solution tho. will change this

edbeeching:
Hi @andimarafioti , I think the slurm scripts still reference the conda envs and should have been updated in this PR.
Also, I am not sure what the advantage of using uv as described in the README, as we are still using pip to install the deps?

I will make a PR to fix.

PR #92, opened 2025-01-28
Improve repoduction of r1 reported score

hynky1999:
- bumps lighteval to the newest commit
- enforces matching boxed environment first when searching for the answer as the r1 models tend to output this as their final answer.

PS: forgot to swtich branch, but since we squash it doesn't matter
PS: don't merge yet

PR #93, opened 2025-01-28
Adds MATH-500 eval results and launch commands

edbeeching:
Adds MATH-500 eval results and launch commands. Requires #92 to be merged


lewtun (approved):
LGTM!

PR #103, opened 2025-01-29
Fix help text for --retries to match actual default value

ATaylorAerospace:
The argparse help text for --retries stated the default was 3, but the actual default set in code is 0. This update corrects the help text to prevent confusion.


lewtun (approved):
Thanks for the fix!

ATaylorAerospace:
No problem

PR #134, opened 2025-01-31
Update grpo.py

wzy6642:
**Improved Exception Handling:** Exceptions are caught and the reward is set to 0 to ensure the code continues to run even if an unexpected error occurs.


qgallouedec:
Duplicate #158

PR #144, opened 2025-01-31
Refactoring reward functions. Adding step by step reasoning reward. Adding test coverage for reward functions

zeenolife:
I've placed the reward functions into their own file to accommodate for future expansion (code specific reward functions and etc.). Added tests to make sure the reward functions are working as intended. 
Also, I've written an optional step by step thinking encouragement function

### DIFF
--- a/src/open_r1/grpo.py
+++ b/src/open_r1/grpo.py
@@ -25,11 +25,10 @@ from datasets import load_dataset
 from transformers import set_seed
 from transformers.trainer_utils import get_last_checkpoint
 
-from latex2sympy2_extended import NormalizationConfig
-from math_verify import LatexExtractionConfig, parse, verify
 from open_r1.configs import GRPOConfig
 from open_r1.utils.callbacks import get_callbacks
 from trl import GRPOTrainer, ModelConfig, ScriptArguments, TrlParser, get_peft_config
+from open_r1.rewards import REWARD_FUNCS_REGISTRY
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/open_r1/grpo.py
```diff
@@ -41,66 +40,17 @@ class GRPOScriptArguments(ScriptArguments):
 
     Args:
         reward_funcs (`list[str]`):
-            List of reward functions. Possible values: 'accuracy', 'format'.
+            List of reward functions. Possible values are dynamically populated from REWARD_FUNCS_REGISTRY.
     """
 
     reward_funcs: list[str] = field(
         default_factory=lambda: ["accuracy", "format"],
```
zeenolife, line 47: I've left the default same

src/open_r1/rewards.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Magic nubmer 3 to encourage 3 steps and more, otherwise partial reward
+    return [min(1.0, count / 3) for count in matches]
+
+
+REWARD_FUNCS_REGISTRY = {
+    "accuracy": accuracy_reward,
+    "format": format_reward,
+    "reasoning_steps": reasoning_steps_reward,
+}
```
zeenolife, line 79: Not exactly fan of this "registry", but list of reward functions is small, so I guess it works for now ¯\_(ツ)_/¯

src/open_r1/rewards.py
```diff
@@ -0,0 +1,79 @@
...
+                        try_extract_without_anchor=False,
+                    )
+                ],
+                extraction_mode="first_match",
+            )
+            # Reward 1 if the content is the same as the ground truth, 0 otherwise
+            reward = float(verify(answer_parsed, gold_parsed))
+        else:
+            # If the gold solution is not parseable, we reward 1 to skip this example
+            reward = 1.0
```
ocramz, line 43: I don't follow: why do we give the same reward in the "non parseable" case and the correct case?

zeenolife: The function was taken as is: [link](https://github.com/huggingface/open-r1/pull/144/files#diff-aa2b823b9a08fc45ead8f8707f0b0ee95f40940d85883ee1fe7edadf8a4c2b0bL83)

But, I speculate that:

1. It's rare that the ground truth dataset is not parseable.
2. The downstream GRPO trainer currently doesn't support ignoring the samples.

ocramz: shouldn't we give it 0 weight instead then?

qgallouedec: It's completely equivalent, the reward is normalized per group.

zeenolife: Yes, I was just thinking of it. 

I guess it is guaranteed that group would contain only the completions for that particular unparseable solution, right? Because if the group would contain others, the normalization wouldnt work properly. Does it make sense?

zeenolife: Nevertheless, I can come back to it in the next PR

ctjlewis: I'd just made an issue about this and was reading this thread to see if it was related. I thought it was odd because the odds were like 0.19, 0.20, etc., so rewarding 1 for a pair where we couldn't find the gold solution felt off.

https://github.com/huggingface/open-r1/issues/159

tests/test_rewards.py
```diff
@@ -0,0 +1,91 @@
+import unittest
+from open_r1.rewards import accuracy_reward, format_reward, reasoning_steps_reward
+
+
+class TestRewards(unittest.TestCase):
+    def test_accuracy_reward_correct_answer(self):
+        """Test accuracy_reward with a correct answer."""
+        completion = [[{"content": r"\boxed{\frac{63}{400}}"}]]
```
winglian, line 8: shouldn't the test be without the `\boxed` ? nowhere in the system prompt do we tell the model to use `\boxed` so even if it gets the answer right with say `<answer>\frac{63}{400}</answer>`, shouldn't that be counted as correct?

zeenolife: Valid point, it's passed in as system prompt argument

zeenolife:
@kashif @lewtun @edbeeching tagging other maintainers

src/open_r1/rewards.py
qgallouedec: > We wouldn't want to give it 0 weight either since that could be a penalty or even a net reward depending on the other rewards in the group

That's not right. Gold is shared in the group. If it's not parsable for one sample, then it's not parsable for the all group.

ctjlewis: Oh OK, I see now.

edbeeching (approved):
LGTM, thanks for this refactor.

zeenolife:
Rebased because of the merge conflict

kashif (approved):

qgallouedec (approved):

PR #149, opened 2025-02-01
Update grpo.py with working imports

saidineshpola:
Fix for wrong imports of open_r1 instead of src.open_r1

replace open_r1 imports with 
`from src.open_r1.configs import GRPOConfig`
`from src.open_r1.utils.callbacks import get_callback`


edbeeching:
@saidineshpola thanks for your PR but I think you have not installed the package correctly, please follow the instructions in the README.

PR #157, opened 2025-02-02
Allow Whitespaces Between XML Tags for Format Reward

andyl98:
As the title suggested.

One common pattern I've observed with the existing system message is that, smaller models (<= 7b) like to produce extra whitespace or newlines in between the `<think> .. </think>` and `<answer> .. </answer>` blocks.

In the system prompt, although there's one-shot example, it is not 100% to the LLM that no space is allowed in between these tags.

This one-line fix will allow the smaller models to quickly get this reward without a lot of trial and errors, essentially speeding up the convergence imo.


lewtun (approved):
Thanks for the improvement! The change seems reasonable to me - WDYT @qgallouedec ?

qgallouedec (approved):
Looks good, thanks

lewtun:
@andyl98 would you mind resolving the merge conflicts and then we can merge this?

andyl98:
Oh seems you guys already have a fix for this. Closing it

qgallouedec:
Actually I think we don't have the `DOTALL`

andyl98:
Prev:
```
    -pattern = r"^<think>.*?</think>\s*<answer>.*?</answer>$"
    -matches = [re.match(pattern, content, re.DOTALL | re.MULTILINE) for content in completion_contents]
```
Now:
```python
   +matches = [re.match(pattern, content) for content in completion_contents]
   +pattern = re.compile(r"^<think>.*?</think>\s*<answer>.*?</answer>$", re.DOTALL)
```

I think the \s* part should have already captured the corner cases

PR #179, opened 2025-02-04
Puzzles updates

Rocketknight1:


Rocketknight1:
Closing for now since we're not sure if `puzzles` will be that helpful or impactful yet. Might reopen in future if we decide we need an extensible task library like that!

PR #181, opened 2025-02-04
chore(setup.py): bump vllm>=0.7.1

ctjlewis:
See https://github.com/huggingface/trl/pull/2766.


kashif:
updated in trl main

lewtun (approved):
Thanks for the fix! @kashif I realise that `vllm` is pinned in `trl`, but it's also good IMO to pin here because you need to install it before the other dependencies (to get the right `torch` binaries)

Issue #145, opened 2025-01-31
AttributeError: 'Qwen2ForCausalLM' object has no attribute 'optimizer' during DRPO training


Some-random:
Faced this error when running evaluation with the latest code. The command to replicate: 
`ACCELERATE_LOG_LEVEL=info accelerate launch --config_file recipes/accelerate_configs/zero3.yaml --num_processes=7 src/open_r1/grpo.py --config recipes/qwen/Qwen2.5-1.5B-Instruct/grpo/confg_full.yaml
`

And the error:

```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/fsx/users/dongweij/open-r1/src/open_r1/grpo.py", line 237, in <module>
[rank0]:     main(script_args, training_args, model_args)
[rank0]:   File "/fsx/users/dongweij/open-r1/src/open_r1/grpo.py", line 189, in main
[rank0]:     train_result = trainer.train(resume_from_checkpoint=checkpoint)
[rank0]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/fsx/users/dongweij/miniconda3/envs/openr1/lib/python3.11/site-packages/transformers/trainer.py", line 2175, in train
[rank0]:     return inner_training_loop(
[rank0]:            ^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/fsx/users/dongweij/miniconda3/envs/openr1/lib/python3.11/site-packages/transformers/trainer.py", line 2554, in _inner_training_loop
[rank0]:     self._maybe_log_save_evaluate(
[rank0]:   File "/fsx/users/dongweij/miniconda3/envs/openr1/lib/python3.11/site-packages/transformers/trainer.py", line 3027, in _maybe_log_save_evaluate
[rank0]:     metrics = self._evaluate(trial, ignore_keys_for_eval)
[rank0]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/fsx/users/dongweij/miniconda3/envs/openr1/lib/python3.11/site-packages/transformers/trainer.py", line 2981, in _evaluate
[rank0]:     metrics = self.evaluate(ignore_keys=ignore_keys_for_eval)
[rank0]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #146, opened 2025-02-01
[GRPO Training Errors] 1.(Qwen2ForCausalLM object has no attribute 'optimizer' ); 2.(ZeroDivisionError)


Niemand03:
```yaml
# /recipes/qwen/Qwen2.5-1.5B-Instruct/grpo/confg_full.yaml

# Model arguments
model_name_or_path: deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B
model_revision: main
torch_dtype: bfloat16

# Data training arguments
dataset_name: AI-MO/NuminaMath-TIR
dataset_configs:
- all
# Num processes is less by 1 as vLLM is using 1 GPU
num_processes: 7

# GRPO trainer config
bf16: true
use_vllm: true
vllm_device: cuda:0
vllm_gpu_memory_utilization: 0.7
do_eval: true
eval_strategy: steps
eval_steps: 100
gradient_accumulation_steps: 16
gradient_checkpointing: true
gradient_checkpointing_kwargs:
  use_reentrant: false
hub_model_id: Qwen2.5-1.5B-Open-R1-GRPO
hub_strategy: every_save
learning_rate: 2.0e-05
log_level: info
logging_steps: 10
logging_strategy: steps
lr_scheduler_type: cosine
max_prompt_length: 512
max_completion_length: 1024
max_steps: -1
num_train_epochs: 1
output_dir: data/Qwen2.5-1.5B-Open-R1-GRPO
overwrite_output_dir: true
per_device_eval_batch_size: 4   
per_device_train_batch_size: 1
push_to_hub: true
report_to:
- wandb
save_strategy: "no"
seed: 42
warmup_ratio: 0.1
```

```yaml
# /recipes/accelerate_configs/zero3.yaml

compute_environment: LOCAL_MACHINE
debug: false
deepspeed_config:
  deepspeed_multinode_launcher: standard
  offload_optimizer_device: none
  offload_param_device: none
  zero3_init_flag: true
  zero3_save_16bit_model: true
  zero_stage: 3
distributed_type: DEEPSPEED
downcast_bf16: 'no'
machine_rank: 0
main_training_function: main
mixed_precision: bf16
num_machines: 1
num_processes: 8
rdzv_backend: static
same_network: true
tpu_env: []
tpu_use_cluster: false
tpu_use_sudo: false
use_cpu: false
```

### Command to Run
```bash
[...]

PR #197, opened 2025-02-05
Provide a minimal reproducible experiment using GRPO for mathematical reasoning on base model, referencing the approach from SimpleRL-Reason

Some-random:
Also solve the [issue](https://github.com/huggingface/open-r1/issues/145) referenced here. We should release a stable recipe that can run end-to-end, even with a small model and dataset? This would help users better understand the workflow and facilitate learning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ To train via the GRPO trainer, we use one GPU to run vLLM for faster generation 
 ACCELERATE_LOG_LEVEL=info accelerate launch --config_file recipes/accelerate_configs/zero3.yaml --num_processes=7 src/open_r1/grpo.py --config recipes/qwen/Qwen2.5-1.5B-Instruct/grpo/confg_full.yaml
 ```
 
-We provide a minimal reproducible experiment using GRPO for mathematical reasoning, referencing the approach from SimpleRL-Reason which uses a 7B model trained on 8K examples. Running this on 8 H100 80G GPU takes about 3 hours:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -119,6 +119,14 @@ To train via the GRPO trainer, we use one GPU to run vLLM for faster generation
 ACCELERATE_LOG_LEVEL=info accelerate launch --config_file recipes/accelerate_configs/zero3.yaml --num_processes=7 src/open_r1/grpo.py --config recipes/qwen/Qwen2.5-1.5B-Instruct/grpo/confg_full.yaml
 ```
 
+We provide a minimal reproducible experiment using GRPO for mathematical reasoning, referencing the approach from [SimpleRL-Reason](https://hkust-nlp.notion.site/simplerl-reason) which uses a 7B model trained on 8K examples. Running this on 8 H100 80G GPU takes about 3 hours:
+
+```shell
+ACCELERATE_LOG_LEVEL=info accelerate launch --config_file recipes/accelerate_configs/zero2.yaml --num_processes=7 src/open_r1/grpo.py --config recipes/deepseek/DeepSeek-R1-Distill-Qwen-7B/grpo/config_base_math_smalllr.yaml
```
edbeeching, line 125: I am curious why you went with zero2 and not zero3?

edbeeching (approved):

Some-random:
Because people have faced [with]([url](https://github.com/huggingface/open-r1/issues/146)) [errors](https://github.com/huggingface/open-r1/issues/145) while using zero3

ctjlewis:
@Some-random @edbeeching, I have reproducible setup and zero3 tested on clean Debian with CUDA 12.1, vllm, 8x H100 in #199. It needed specific package versions pinned.

I have to update branch though. I could test this one for zero3 if we want, after branch update.

PR #209, opened 2025-02-06
Update CUDA

lewtun:


Ryan-ly:
Can I know the reason for upgrading CUDA from version 12.1 to 12.4? We are currently reproducing openr1, but we have already installed CUDA 12.1 and the corresponding nvidia driver.  Would there be any impact if we don't upgrade CUDA?

Issue #159, opened 2025-02-02
Should skips really get reward 1?


ctjlewis:
For https://github.com/huggingface/open-r1/pull/158, to handle situations where `math_verify` throws an error, we just skip it with existing behavior. 

But what is the reasoning behind awarding reward 1, the same as a correct answer? I'd assume we just return `None` and _actually_ skip it.

ctjlewis:
Looking at this, I really don't think it's correct. 

We are comparing the rewards between for different solutions in a group, and we might give a solution a reward of 1, even if the average we're shooting for that group is negative one million (or they're all just zeroes), and the reward for those should not be the same as a correct answer. We're diluting the signal of the correct answers.

https://github.com/huggingface/trl/blob/a325a0eec5b646f3f87a1284abbc9b9790a4e442/trl/trainer/grpo_trainer.py#L471-L501

Some-random:
Agreed. I also don't think we should give an reward of 1 when math_verify fails to work

qgallouedec:
It doesn't really matter, the rewards are normalized anyway.

qgallouedec:
The only problem is logging. If it turns out that half of your dataset has a non-parsable solution, then you'll always have a logged reward of at least 50%, which is misleading. But in terms of optimization, it doesn't make any difference, because rewards are normalized per group.

Ideally, we'd like to be able to remove non-parsable solution samples, or more generally, to return None when the reward function doesn't apply to the sample for whatever reason, but this requires some changes in the way the advantage is counted on the TRL side. We could add this feature in the future, but for now, it's an edge case. Who works with a 50% non parsable dataset?

ctjlewis:
@qgallouedec Please see this thread: https://github.com/huggingface/open-r1/pull/144#discussion_r1941430737

How does normalizing get us away from this? We are introducing signals that aren't there. My rewards are [0.1, 0.1, 0.1, 0.1], and then the skip is reward: 1. That is not desirable. We cannot do reward 0 either.

ctjlewis:
It is technically a bugfix and not a feature. We do not want to reward an output because it happened to be skipped.

ctjlewis:
OK, if it fails once it fails for the whole group, so this is not relevant, closing.

PR #213, opened 2025-02-07
Weighted reward functions

zeenolife:
It doesn't feel right that "easy" reward for formatting weighs the same as getting the right answer for a "hard" math question. In fact, the Unsloth have been using varied weights for GRPO training.
So:
- Adding functionality to weigh rewards. 
- Adding some tests.


zeenolife:
PTAL @kashif @edbeeching

recipes/deepseek/DeepSeek-R1-Distill-Qwen-7B/grpo/config_base_math_smalllr.yaml
```diff
@@ -40,6 +40,12 @@ per_device_train_batch_size: 1
 push_to_hub: true
 report_to:
 - wandb
+reward_funcs:
+  - accuracy
+  - format
+reward_weights:
+  - 1.0
+  - 0.5
```
edbeeching, line 48:
```suggestion
  - 1.0
```
While I agree with the weights you proposed, can we have the values in the recipes match the original values, otherwise we would be unable to reproduce their exact results.

recipes/deepseek/DeepSeek-R1-Distill-Qwen-7B/grpo/config_full.yaml
```diff
@@ -40,6 +40,12 @@ per_device_train_batch_size: 1
 push_to_hub: true
 report_to:
 - wandb
+reward_funcs:
+  - accuracy
+  - format
+reward_weights:
+  - 1.0
+  - 0.5
```
edbeeching, line 48:
```suggestion
  - 1.0
```

recipes/qwen/Qwen2.5-1.5B-Instruct/grpo/confg_full.yaml
```diff
@@ -40,6 +40,12 @@ per_device_train_batch_size: 1
 push_to_hub: true
 report_to:
 - wandb
+reward_funcs:
+  - accuracy
+  - format
+reward_weights:
+  - 1.0
+  - 0.5
```
edbeeching, line 48:
```suggestion
  - 1.0
```

kashif:
@qgallouedec i can fix the merge conflicts as its due to my stuff

edbeeching (changes requested):
LGTM with minor changes

src/open_r1/grpo.py
```diff
@@ -41,6 +41,8 @@ class GRPOScriptArguments(ScriptArguments):
     Args:
         reward_funcs (`list[str]`):
             List of reward functions. Possible values are dynamically populated from REWARD_FUNCS_REGISTRY.
+        reward_weights (`list[float]`):
```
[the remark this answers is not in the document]
zeenolife: They should, let me fix it, and rebase it

zeenolife:
> I think there's a misunderstanding here about how GRPO works:
> 
> Applying a coefficient doesn't change the optimization because rewards are normalized _by group_ and not globally. The only way to do this is to change the way rewards are aggregated, as here: [huggingface/trl#2676](https://github.com/huggingface/trl/pull/2676)

Could you please elaborate what do you mean by group?
AFAIU, it’s the multiple completions for the **same** question. 

Now say there are 3 completions: A, B and C. 
All of them go through the list of reward functions e.g [accuracy, format]
Let’s say A got accuracy right, B got format right and C got both right. 

Previously:
A: 1 point
B: 1 point
C: 2 points
Mean: 1.33
A pseudo  normalized: -0.33
B pseudo normalized: -0.33
C pseudo normalized: 0.66

Now:
A: 1 point
B: 0.5 point
C: 1.5 points
Mean: 1
A pseudo  normalized: 0
B pseudo normalized: -0.5
C pseudo normalized: 0.5


Do I have the right picture?

qgallouedec:
If a user sets `weights = [1, 2]`, I bet they expects reward B to count for twice as much as reward A. But it's not what the code is doing here

zeenolife:
> rewards are normalized globally
I see, so that means that the logic around the accuracy function has to be changed to not penalize/reward absense of gold_parsed compared to other questions` completions.

Where do we stand on this PR?

> it's very counter-intuitive because by reward_weights, I (and probably everyone else) expect that behind, there's a weighted average between the rewards.

I can see the confusion here, but I think the term "weight" is subtly different than "coefficient", where the former is more of a notion of relative importance, whereas latter is simple multiplier. But I don't have a strong opinion on this.

> For example, one consequence of doing it this way is that if your reward function is centered in 0, applying a coefficient won't have any impact.

But it does, as I've shown above. The behaviour it changes is relative to each reward function in reward system, not relative to different completions. It is the completions that are centered around 0, not reward functions.


We could change the naming to `reward_relative_weight` the remove ambiguity.
But I strongly believe that giving correct answer higher importance compared to simple formatting is very much needed

ctjlewis:
> Forget everything I've said, I've been wrong all along! rewards are normalized globally

- https://github.com/huggingface/open-r1/issues/159
 
Wasn't that the basis for not touching this, do we want to reopen or is this still not an issue? My concern is just that reward 1 might be very high depending on the reward function.

qgallouedec:
I propose the following:

What is proposed here seems to me to make more sense on the TRL side. And it avoids adding complexity to Open-R1. Nevertheless, we can (1) merge this in the meantime. (2) integrate the functionality into TRL (3) remove what can be removed from Open-R1.
/

qgallouedec:
> My concern is just that reward 1 might be very high (or low) (for no reason) depending on the reward function. So even if everyone in the group has equal advantage, that group's mean reward will still always be exactly 1.

I don't think so; let's check the two cases

| completion idx | correctness | format | total reward | Advantage |
|-|-|-|-|-|
| **0** | 1.0 | 1.0 | 2.0 | 0.66 |
| **1** | 1.0 | 0.0 | 1.0 | -0.33 |
| **1** | 1.0 | 0.0 | 1.0 | -0.33 |

| completion idx | correctness | format | total reward | Advantage  |
|-|-|-|-|-|
| **0** | 0.0 | 1.0 | 1.0 | 0.66 |
| **1** | 0.0 | 0.0 | 0.0 | -0.33 |
| **1** | 0.0 | 0.0 | 0.0 | -0.33 |

In both cases you get the same advantages, right?

> Ideally it would just not come into the training run at all, there is no signal to be gained when we can't parse gold.

Agree, do you want to integrate it in TRL? It's basically supporting reward function that can return None

src/open_r1/rewards.py
```diff
@@ -197,3 +197,21 @@ def repetition_penalty_reward(completions, **kwargs) -> float:
...
+
+    Args:
+        func: The reward function to weight
+        weight: The weight to apply to the reward
+
+    Returns:
+        A new function that applies the weight to the reward
+    """
+
+    def weighted_reward(*args, **kwargs):
```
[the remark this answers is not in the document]
zeenolife: Oh, that's a good point, it probably would, because I'm creating new function. Let me look into this a bit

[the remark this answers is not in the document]
zeenolife: A quick fix could be introducing weight argument to each reward function instead. I don't exactly like it though. What do you think?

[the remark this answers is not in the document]
zeenolife: Nvm, I've put functools.wraps decorator to preserve the metadata, and added test that checks the `__name__` remains the same

qgallouedec (approved):
Smart, thanks for iterating @zeenolife

zeenolife:
Ah, looks like tests are failing because trl expects to have CUDA device.

Do we intend to upgrade the CI to run on GPU machine?

As an option I can:
- Disable GRPO tests entirely for now
- Run the tests if CUDA device is available

src/open_r1/grpo.py
```diff
@@ -57,6 +61,7 @@ class GRPOScriptArguments(ScriptArguments):
             Maximum reward for cosine scaling for correct answers.
         cosine_max_len (`int`):
             Maximum length for cosine scaling.
+
```
qgallouedec, line 64:
```suggestion
```

tests/test_rewards.py
```diff
@@ -110,6 +140,17 @@ def test_format_reward_specific_multiline(self):
         rewards = format_reward(completion)
         self.assertEqual(rewards[0], 1.0)
 
+    def test_weighted_reward_preserves_name(self):
+        """Test that create_weighted_reward preserves the original function name."""
```
qgallouedec, line 144:
```suggestion
        """Test that create_weighted_reward preserves the original function name. Important for logging."""
```

qgallouedec:
> h, looks like tests are failing because trl expects to have CUDA device.

it doesn't actually, I'm quite surprise, let me dig in

ctjlewis:
> Agree, do you want to integrate it in TRL? It's basically supporting reward function that can return None

@qgallouedec I went to do this but couldn't get grpo_trainer tests to run even with 8x H100. Is that fixed on this branch? The first test passed but everything else failed.

PR #229, opened 2025-02-07
Fix `cosine_scaled_reward` compatibility with GRPO

qgallouedec:


src/open_r1/grpo.py
```diff
@@ -55,7 +55,7 @@ class GRPOScriptArguments(ScriptArguments):
     """
 
     reward_funcs: list[str] = field(
-        default_factory=lambda: ["accuracy", "format", "reasoning_steps", "cosine"],
+        default_factory=lambda: ["accuracy", "format"],
```
qgallouedec, line 58:
```suggestion
        default_factory=lambda: ["accuracy", "format", "reasoning_steps", "cosine"],
```

src/open_r1/grpo.py
```diff
@@ -26,7 +26,7 @@
 from transformers.trainer_utils import get_last_checkpoint
 
 from open_r1.configs import GRPOConfig
-from open_r1.rewards import accuracy_reward, cosine_scaled_reward, format_reward, reasoning_steps_reward
+from open_r1.rewards import accuracy_reward, format_reward, get_cosine_scaled_reward, reasoning_steps_reward
```
lewtun, line 29: turbo nit: given the other reward functions aren't getters, what about we call this `cosine_scaled_reward` and use`_cosine_scaled_reward` as the inner function

lewtun (approved):
Thanks for the fix! LGTM with a small question about replacing the getter with the same naming convention as other reward functions

src/open_r1/grpo.py
qgallouedec: I think I would be more confusing actually, because you'd have this:

```python
REWARD_FUNCS_REGISTRY = {
    "accuracy": accuracy_reward,
    "format": format_reward,
    "reasoning_steps": reasoning_steps_reward,
    "cosine": cosine_scaled_reward(min_value_wrong=min_value_wrong, ...)
    ),
}
```

and in the logged metrics:

```{..., 'rewards/accuracy_reward': 0.6, 'rewards/format_reward': 0.0, 'rewards/_cosine_scaled_reward': 0.0, 'epoch': 0.0}```

lewtun: ah good point. let's keep the getter then

PR #231, opened 2025-02-07
Refactor training configs and unify Slurm for training SFT & GRPO

lewtun:
Cleans up how we do training so we just need a single `train.slurm` script. I've also add FA2 as a core dependency and updated the various configs.


edbeeching (approved):
LGTM, by they way did you test with `OPTIONAL_ARGS` ? As the parser works differently from the H4 one.

lewtun:
> LGTM, by they way did you test with `OPTIONAL_ARGS` ? As the parser works differently from the H4 one.

Ah good point, let me check it

lewtun:
`OPTIONAL_ARGS` works 👌

Julian-Schaefer:
Screenshot from 2025-02-08 13-00-44

I can't install the dependencies anymore after this update because of flash attention. I freshly cloned the repo and followed the installation instructions. I am also using Cuda 12.4, but I still can't get it to install

PR #232, opened 2025-02-07
Use new GRPO logic

qgallouedec:
From https://github.com/huggingface/trl/commit/cf97133d51e2e0d9f945fd8502064e304288bb54, use this new config.

### DIFF
--- a/recipes/deepseek/DeepSeek-R1-Distill-Qwen-7B/grpo/config_base_math_smalllr.yaml
+++ b/recipes/deepseek/DeepSeek-R1-Distill-Qwen-7B/grpo/config_base_math_smalllr.yaml
@@ -32,11 +32,12 @@ lr_scheduler_type: cosine
 max_prompt_length: 512
 max_completion_length: 1024
 max_steps: -1
+num_generations: 7
 num_train_epochs: 1
 output_dir: data/Qwen-2.5-7B_Base_Math_smalllr
 overwrite_output_dir: true
-per_device_eval_batch_size: 1
-per_device_train_batch_size: 1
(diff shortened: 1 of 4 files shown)


recipes/deepseek/DeepSeek-R1-Distill-Qwen-7B/grpo/config_base_math_smalllr.yaml
```diff
@@ -32,11 +32,12 @@ lr_scheduler_type: cosine
 max_prompt_length: 512
 max_completion_length: 1024
 max_steps: -1
+num_generations: 7
```
qgallouedec, line 35: implicitly implies that we use 7 training GPUs. Not ideal but any simple alternative?

edbeeching: Seems ok given the improvements in memory usage.
FYI I ran their original config yesterday if you want to compare running the new recipe against it: 
https://wandb.ai/edbeeching/huggingface/runs/p3otw3m9?nw=nwuseredbeeching
(let me know if you cannot access the wandb report)

qgallouedec: cool, thanks!

lewtun:
Ah sorry for the big refactor I did in #231 🙈 

Your changes makes sense, so feel free to merge after adapting to the new config paths (might be faster to close the PR and open a new one)

Issue #237, opened 2025-02-08
Format reward problem


JamesHujy:
The format reward seems to have some problems.

I tested the current reward function using the following code. The input is a model-generated output after training for 100 steps:  

```python
import re

inputs = "<think>\nI will count each distinct object in the image:\n1. Purple scooter\n2. Red bicycle\n3. Green motorcycle\n4. Gray sedan\n5. Yellow school bus\n6. Small green double-decker bus\n7. Small red car\n8. Small purple car\n9. Small gray dirt bike\n\nThere are 9 distinct objects in total.\n</think>\n<answer>9</answer>"

def format_reward(completions):
    """Reward function that checks if the completion follows a specific format."""
    pattern = r"<think>.*?</think>\s*<answer>.*?</answer>"
    matches = [re.match(pattern, content) for content in completions]
    return [1.0 if match else 0.0 for match in matches]
```

However, this function incorrectly returns `0` for valid inputs.  

The issue occurs because the `.*?` wildcard does not match newlines unless `re.DOTALL` is enabled.  Updating the regex as follows ensures that the entire string matches the expected format:  

```python
pattern = r"<think>.*?</think>\s*<answer>.*?</answer>"
matches = [re.match(pattern, content, re.DOTALL) for content in completions]
```

I valid this format reward in the VLM setting with Qwen2-VL. See https://github.com/Deep-Agent/R1-V/issues/19

asirgogogo:
great！！！

brishtiteveja:
Related node: Please also check whether you are using cuda 12.4 (as it is shown as Caution in ReadMe). I was initially using cuda 12.1 and then switched to 12.4

qgallouedec:
Closed by #238

PR #238, opened 2025-02-08
fix format reward

JamesHujy:
See https://github.com/huggingface/open-r1/issues/237

### DIFF
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -51,9 +51,9 @@ def accuracy_reward(completions, solution, **kwargs):
 
 def format_reward(completions, **kwargs):
     """Reward function that checks if the completion has a specific format."""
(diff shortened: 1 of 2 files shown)


lewtun (commented):
Great catch @JamesHujy and thank you for the fix! Would you mind adding your example from #237 as a regression test in `/tests`? Then we should be good to merge!

kashif:
@JamesHujy @lewtun so the current regex still fails for the input example from #237 let me fix it

kashif:
we should add a `\s*` between the `</think>` and `<answer>` to handle white space or newlines and also perhaps add a `re.MULTILINE` in addition to the `re.DOTALL` for multiline inputs

lewtun (approved):
Thanks for the fixes @kashif !

Issue #163, opened 2025-02-02
Error in make evaluate: TypeError: VLLMModelConfig.__init__() got an unexpected keyword argument ''


mattdepaolis:
When running the `make evaluate` command with the specified parameters, an error occurs indicating that the `VLLMModelConfig` initializer received an unexpected keyword argument.

**Steps to Reproduce:**

1. Set up the environment as per the project's guidelines.
2. Execute the following command:
```bash
make evaluate MODEL=deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B TASK=math_500
```

**Expected Behavior**: The evaluation process should complete without errors, utilizing the specified model and task parameters.

**Actual Behavior:** The following error message is displayed:
```css
TypeError: VLLMModelConfig.__init__() got an unexpected keyword argument ''
```
**Analysis:** Upon investigation, it appears that the `MODEL_ARGS` variable in the `Makefile`'s `evaluate` target may include an empty `$(PARALLEL_ARGS)` variable, leading to consecutive commas and resulting in an empty string being passed as a keyword argument.

Proposed Solution: Modify the `MODEL_ARGS` assignment in the `Makefile` to conditionally include `$(PARALLEL_ARGS)` only if it's not empty:
```makefile
MODEL_ARGS="pretrained=$(MODEL),dtype=float16$(if $(PARALLEL_ARGS),,$(PARALLEL_ARGS)),max_model_length=32768,gpu_memory_utilisation=0.8"
```
This change ensures that no extra comma is introduced when `$(PARALLEL_ARGS)` is empty, preventing the unintended empty string argument.

kevinss30:
Try to replace `gpu_memory_utilisation` with `gpu_memory_utilization`. It's a argument in vllm which may wrongly documented in readme.

PR #246, opened 2025-02-08
Fix: Avoid empty keyword argument in VLLMModelConfig from Makefile

mattdepaolis:
**Issue Summary**
When running the make evaluate command with the specified parameters, an error occurs indicating that the VLLMModelConfig initializer received an unexpected keyword argument.

- Fixes : [Error in make evaluate: TypeError: VLLMModelConfig.__init__() got an unexpected keyword argument '' #163](https://github.com/huggingface/open-r1/issues/163)

**Steps to Reproduce:**

1. Set up the environment as per the project's guidelines.
2. Execute the following command:

```bash
make evaluate MODEL=deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B TASK=math_500
```

3. Expected Behavior: The evaluation process should complete without errors, utilizing the specified model and task parameters.

4. Actual Behavior: The following error message is displayed:

```css
TypeError: VLLMModelConfig.__init__() got an unexpected keyword argument ''
```

**Root Cause Analysis**
- The MODEL_ARGS variable in the Makefile's evaluate target includes an empty $(PARALLEL_ARGS), which introduces consecutive commas.
- This results in an empty string being passed as a keyword argument to VLLMModelConfig.

**Proposed Solution**
Modify the `MODEL_ARGS` assignment in the Makefile to conditionally include `$(PARALLEL_ARGS)` only if it's not empty:

```make
MODEL_ARGS="pretrained=$(MODEL),dtype=bfloat16$(if $(PARALLEL_ARGS),,$(PARALLEL_ARGS)),max_model_length=32768,gpu_memory_utilisation=0.8"
```

This prevents the introduction of an unintended empty string argument.

**Testing**
- Ran the make evaluate command with and without PARALLEL_ARGS and verified that no unexpected keyword argument error occurs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ evaluate:
 		fi \
 	),))
 	$(if $(filter tensor,$(PARALLEL)),export VLLM_WORKER_MULTIPROC_METHOD=spawn &&,) \
-	MODEL_ARGS="pretrained=$(MODEL),dtype=bfloat16,$(PARALLEL_ARGS),max_model_length=32768,gpu_memory_utilisation=0.8" && \
+	MODEL_ARGS="pretrained=$(MODEL),dtype=bfloat16$(if $(PARALLEL_ARGS),,$(PARALLEL_ARGS)),max_model_length=32768,gpu_memory_utilisation=0.8" && \
 	lighteval vllm $$MODEL_ARGS "custom|$(TASK)|0|0" \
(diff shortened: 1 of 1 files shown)


Charlesnorris509 (commented):
@mattdepaolis This PR seems to be solid and follow best practices, however The **evaluate** target is quite complex and involves multiple nested conditionals and shell commands. This will be difficult to  maintain.

PR #247, opened 2025-02-08
Fix README: Correct recipes path and missing --config option

tyfeng1997:
This PR fixes two issues in the training command in **README.md**:

1. Incorrect path: Adjusted the path to match the actual folder structure.
2. Missing --config flag: Added --config before the YAML file.


lewtun (approved):
Thanks for the fix!

PR #262, opened 2025-02-10
Initial GRPO exps on the Numina dataset

edbeeching:
Adding initial GRPO exps on the numina for the `deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B` model

### DIFF
--- a/recipes/DeepSeek-R1-Distill-Qwen-1.5B/grpo/config_v00.00.yaml
+++ b/recipes/DeepSeek-R1-Distill-Qwen-1.5B/grpo/config_v00.00.yaml
@@ -1,0 +1,53 @@
+# Model arguments
+model_name_or_path: deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B
+model_revision: main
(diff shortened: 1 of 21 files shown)


recipes/DeepSeek-R1-Distill-Qwen-1.5B/grpo/config_v00.00.yaml
lewtun, line 1: Could we collapse these files into a single config and share a README with the Slurm commands that override them? That way we have 1 file linked to 1 dataset

Also, WDYT about having something like `config_numina_math.yaml` instead of `config_v00.00` which is a bit cryptic for the community?

lewtun (commented):
LGTM with a comment about squashing the configs into 1 and potentially renaming the config to make it easier for others to understand which one is which (just ideas, happy to go with what you prefer)

edbeeching:
closing, as we are working on other exps that are more successful and I would prefer not to confuse members of the community

PR #263, opened 2025-02-10
Adds repetition penalty reward

edbeeching:
Adds repetition penalty reward from [Demystifying Long Chain-of-Thought Reasoning in LLMs](https://arxiv.org/abs/2502.03373)

### DIFF
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -155,26 +155,26 @@ def get_repetition_penalty_reward(ngram_size: int, max_penalty: float):
         raise ValueError(f"max_penalty {max_penalty} should not be positive")
 
     if max_penalty == 0:
-        return 0    
-    
+        return 0
+
     def zipngram(text: str, ngram_size: int):
         words = text.lower().split()
         return zip(*[words[i:] for i in range(ngram_size)])
-    
+
     def repetition_penalty_reward(completions, *args, **kwargs):
         """
-            reward function the penalizes repetitions
-            ref implementation: https://github.com/eddycmu/demystify-long-cot/blob/release/openrlhf/openrlhf/reward/repetition.py
-            
-            Args:
-                completions: List of model completions
-                solution: List of ground truth solutions
+        reward function the penalizes repetitions
+        ref implementation: https://github.com/eddycmu/demystify-long-cot/blob/release/openrlhf/openrlhf/reward/repetition.py
 
-            This function is parameterized by the following arguments:
-                ngram_size: size of the n-grams
-                max_penalty: Maximum penalty for wrong answers
+        Args:
+            completions: List of model completions
+            solution: List of ground truth solutions
+
+        This function is parameterized by the following arguments:
+            ngram_size: size of the n-grams
+            max_penalty: Maximum penalty for wrong answers
         """
-        
+
         rewards = []
         for completion in completions:
             if completion == "":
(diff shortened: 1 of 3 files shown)


src/open_r1/rewards.py
```diff
@@ -148,3 +148,51 @@ def cosine_scaled_reward(completions, solution, **kwargs):
         return rewards
 
     return cosine_scaled_reward
+
+
+def get_repetition_penalty_reward(ngram_size: int, max_penalty: float):
+    if max_penalty > 0:
```
kashif, line 154:
```suggestion
    if max_penalty < 0:
```

kashif (approved):

src/open_r1/rewards.py
```diff
@@ -148,3 +148,51 @@ def cosine_scaled_reward(completions, solution, **kwargs):
...
+        raise ValueError(f"max_penalty {max_penalty} should not be positive")
+
+    if max_penalty == 0:
+        return 0
+
+    def zipngram(text: str, ngram_size: int):
+        words = text.lower().split()
+        return zip(*[words[i:] for i in range(ngram_size)])
+
+    def repetition_penalty_reward(completions, *args, **kwargs):
```
lewtun, line 164: Since this function is so simple, shall we ditch `args` and `kwargs` in favour of being more explicit? That also means we don't need a docstring :)

src/open_r1/rewards.py
```diff
@@ -148,3 +148,51 @@ def cosine_scaled_reward(completions, solution, **kwargs):
         return rewards
 
     return cosine_scaled_reward
+
+
+def get_repetition_penalty_reward(ngram_size: int, max_penalty: float):
```
lewtun, line 153:
```suggestion
def get_repetition_penalty_reward(ngram_size: int, max_penalty: float) -> float:
    """Computes N-gram repetition penalty as described in Appendix C.2 of https://arxiv.org/abs/2502.03373. Reference implementation from: https://github.com/eddycmu/demystify-long-cot/blob/release/openrlhf/openrlhf/reward/repetition.py"""
```

lewtun (approved):
LGTM with a small change to simplify the function signature if possible

src/open_r1/rewards.py
qgallouedec: no, it returns a callable

src/open_r1/rewards.py
qgallouedec:
```suggestion
    def repetition_penalty_reward(completions, **kwargs):
```

src/open_r1/rewards.py
```diff
@@ -148,3 +148,51 @@ def cosine_scaled_reward(completions, solution, **kwargs):
...
+        return zip(*[words[i:] for i in range(ngram_size)])
+
+    def repetition_penalty_reward(completions, *args, **kwargs):
+        """
+        reward function the penalizes repetitions
+        ref implementation: https://github.com/eddycmu/demystify-long-cot/blob/release/openrlhf/openrlhf/reward/repetition.py
+
+        Args:
+            completions: List of model completions
+            solution: List of ground truth solutions
```
qgallouedec, line 171:
```suggestion
```

`solutions` isn't used

src/open_r1/rewards.py
edbeeching: added ref, this returns a function, so I added a return type to the actual reward fn

src/open_r1/rewards.py
edbeeching: I have kept `**kwargs` as other reward functions use other arguments and the GRPOTrainer call is not aware of the signature of the function it is calling so provides the additional kwargs, regardless of whether they are used or not.

PR #264, opened 2025-02-10
hardcodes num_processes to 7 when using vllm

edbeeching:
Previously we were using the world size, but when we use vllm for generation we only use 7 devices for optimization, and hence require 7 processes. Should be updated if we scale to multi-node.

### DIFF
--- a/slurm/train.slurm
+++ b/slurm/train.slurm
@@ -32,6 +32,12 @@ WORLD_SIZE=$(($NUM_NODES*$GPUS_PER_NODE))
 # Due to conflicts between Accelerate's DeepSpeed configs and Transformers' TrainingArguments, we need to parse the gradient accumulation steps from the config file to ensure they match
 CONFIG_FILE=recipes/$MODEL/$TASK/config_$CONFIG_SUFFIX.yaml
 GRAD_ACC_STEPS=$(grep 'gradient_accumulation_steps' $CONFIG_FILE | awk '{print $2}')
+USE_VLLM=$(grep 'use_vllm:\s*true' $CONFIG_FILE) # Match "use_vllm: true" (with optional whitespace)
+
+if [ -n "$USE_VLLM" ]; then  # Check if USE_VLLM is *not* empty (found)
+    NUM_PROCESSES=7
+else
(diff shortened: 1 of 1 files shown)


slurm/train.slurm
```diff
@@ -32,6 +32,12 @@ WORLD_SIZE=$(($NUM_NODES*$GPUS_PER_NODE))
 # Due to conflicts between Accelerate's DeepSpeed configs and Transformers' TrainingArguments, we need to parse the gradient accumulation steps from the config file to ensure they match
 CONFIG_FILE=recipes/$MODEL/$TASK/config_$CONFIG_SUFFIX.yaml
 GRAD_ACC_STEPS=$(grep 'gradient_accumulation_steps' $CONFIG_FILE | awk '{print $2}')
+USE_VLLM=$(grep 'use_vllm:\s*true' $CONFIG_FILE) # Match "use_vllm: true" (with optional whitespace)
+
+if [ -n "$USE_VLLM" ]; then  # Check if USE_VLLM is *not* empty (found)
+    NUM_PROCESSES=7
```
lewtun, line 38: Since WORLD_SIZE isn't used anywhere else, how about we do this:

```
if [ -n "$USE_VLLM" ]; then
    WORLD_SIZE=$(($WORLD_SIZE-1))
fi
```

That way it also works if one wants to train on different setups (e.g. 3+1 GPUS)

lewtun (approved):
LGTM with a small suggestion to make it more generic

slurm/train.slurm
```diff
@@ -58,7 +64,7 @@ export LAUNCHER="HF_HUB_ENABLE_HF_TRANSFER=1 ACCELERATE_LOG_LEVEL=info TRANSFORM
     --config_file recipes/accelerate_configs/$ACCELERATOR.yaml  \
     --gradient_accumulation_steps $GRAD_ACC_STEPS \
     --num_machines $NUM_NODES \
-    --num_processes $WORLD_SIZE \
+    --num_processes $NUM_PROCESSES \
```
lewtun, line 67: This should be reverted if you agree with my suggestion above:

```suggestion
    --num_processes $WORLD_SIZE \
```

PR #324, opened 2025-02-14
fix bug: text.split() not token list

hellen9527:
if text = "abcdefg", text.split() will get ["abcdefg"]， need using list(text)


hellen9527:
> I think that in general we do n-grams across words, not letters, so we split on whitespace ` `

I understand now. It's a matter of language differences. In non-English domains, there often aren't natural delimiters, so it's necessary to convert the text into a token list using a list.

Issue #205, opened 2025-02-06
trl/trainer/utils.py", line 612, in <lambda> [rank0]:  self.formatting_func = lambda x: x[dataset_text_field] [rank0]: [rank0]: KeyError: 'text'


bannima:
## SFT 
dataset_name_or_path=bespokelabs/Bespoke-Stratos-17k

accelerate launch --config_file=recipes/accelerate_configs/zero3.yaml src/open_r1/sft.py \
    --model_name_or_path $model_name_or_path \
    --dataset_name $dataset_name_or_path \
    --learning_rate 2.0e-5 \
    --num_train_epochs 1 \
    --packing \
    --max_seq_length 4096 \
    --per_device_train_batch_size 1 \
    --per_device_eval_batch_size 4 \
    --gradient_accumulation_steps 4 \
    --gradient_checkpointing \
    --bf16 \
    --logging_steps 5 \
    --eval_strategy steps \
    --eval_steps 100 \
    --output_dir ckp/Qwen2.5-32B-Open-R1-Distill
## Error
Generating train split: 0 examples [00:00, ? examples/s]
[rank0]: Traceback (most recent call last):
[rank0]:   File "/home/tiger/.local/lib/python3.11/site-packages/datasets/builder.py", line 1606, in _prepare_split_single
[rank0]:     for key, record in generator:
[rank0]:   File "/home/tiger/.local/lib/python3.11/site-packages/datasets/packaged_modules/generator/generator.py", line 33, in _generate_examples
[rank0]:     for idx, ex in enumerate(self.config.generator(**gen_kwargs)):
[rank0]:   File "/home/tiger/.local/lib/python3.11/site-packages/trl/trainer/sft_trainer.py", line 492, in data_generator
[rank0]:     yield from constant_length_iterator
[rank0]:   File "/home/tiger/.local/lib/python3.11/site-packages/trl/trainer/utils.py", line 637, in __iter__
[rank0]:     buffer.append(self.formatting_func(next(iterator)))
[rank0]:                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]:   File "/home/tiger/.local/lib/python3.11/site-packages/trl/trainer/utils.py", line 612, in <lambda>
[rank0]:     self.formatting_func = lambda x: x[dataset_text_field]
[rank0]:                                      ~^^^^^^^^^^^^^^^^^^^^
[rank0]: KeyError: 'text'

[...]

PR #347, opened 2025-02-17
Fix dataset url

Zzhiter:
When I was trying to run this project, I downloaded the dataset from the URL provided in the Readme, but encountered an error during execution, similar to what was described in this issue: [https://github.com/huggingface/open-r1/issues/205 ](https://github.com/huggingface/open-r1/issues/205). 
The solution is provided in the issue, and I believe it needs to be fixed promptly to prevent more people from wasting time on this problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ sudo apt-get install git-lfs
 
 ## Training models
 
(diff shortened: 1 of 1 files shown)


edbeeching (approved):
Thanks LGTM

PR #351, opened 2025-02-17
fix peft config

rohankumar1999:


lewtun:
Thanks for the PR @rohankumar1999 ! Can you explain why we cannot use the `get_peft_config()` helper function?

rohankumar1999:
> Thanks for the PR @rohankumar1999 ! Can you explain why we cannot use the `get_peft_config()` helper function?

wanted to raise this PR on my fork but messed up. anyways, I was running into issues with `get_peft_config()` when I try to give lora related args .

PR #364, opened 2025-02-18
Add E2B code interpreter reward function

lewtun:
This PR adds a reward function to execute Python code safely via E2B: https://e2b.dev/docs/legacy/sandbox/api/debugging

It's currently targeted for coding competitions where the ground truth is given in the form of test cases and the reward per problem is defined as the overall success rate.

Experiments to test the feature can be viewed here: https://wandb.ai/huggingface/open-r1/reports/Qwen2-5-1-5B-Instruct-with-Python-code-interpreter--VmlldzoxMTQxNTA3Mg

## TODO

- [x] Add some docs on how to set this up with E2B
- [x] Run some experiments to sanity check


src/open_r1/rewards.py
```diff
@@ -271,3 +281,80 @@ def repetition_penalty_reward(completions, **kwargs) -> float:
...
+        raise ImportError(
+            "E2B is not available and required for this reward function. Please install E2B with "
+            "`pip install e2b-code-interpreter` and add an API key to a `.env` file."
+        )
+
+    rewards = []
+    # TODO: add support for other languages in E2B: https://e2b.dev/docs/code-interpreting/supported-languages
+    try:
+        """Returns a reward function that evaluates code snippets in a sandbox."""
+        evaluation_script_template = """
```
qgallouedec, line 308: It's surprising that you don't have any issue with the extra indentation.

qgallouedec: I usually use `textwrap.dedent` in this case, but it might not be necessary here for some reason.

qgallouedec (approved):
🔥🔥🔥🔥

PR #376, opened 2025-02-19
[WIP] Adding formal verification API client, GRPO helpers

ocramz:
Introducing a new dataset and family of tasks to reward models that judge code snippets as "correct" in the deductive program verification sense [1][2].

We provide an API endpoint that does random program generation in a toy subset of Python, together with preconditions and postconditions.  Additionally, the API endpoint returns the execution "trace" (how the variable environment is mutated after each statement).

Later, the model under training can be prompted to predict program correctness properties with full or partial trace information.

The API endpoints are parametric, so it's possible to train the model on small ASTs and test on larger ones, or longer programs, etc.

* add GRPO helpers to use the new API endpoints: src/open_r1/rewards/api/code/unfoldml/htgen.py and src/open_r1/rewards/code/htgen.py
* FEEDBACK WANTED: how to prompt a model under test using structured information? Should we use a specific templating format?
* Split "rewards" into a new folder and distinct files. (not sure this was a good idea)

[1] https://en.wikipedia.org/wiki/Correctness_(computer_science)
[2] https://en.wikipedia.org/wiki/Hoare_logic


ocramz:
cc'ing @Muhtasham @vumichien and @lewtun @qgallouedec ^^

PR #377, opened 2025-02-19
Language specific code format reward

zeenolife:
I was looking at the logs of code generation. It does look like sometimes the completion does have valid code, but it doesn't get picked up by `code_reward` function.

This one should replace the format_reward for code gen


zeenolife:
@lewtun please

zeenolife:
Or @kashif perhaps?

PR #392, opened 2025-02-22
Update prompt template and sampling parameters for evaluation

lewtun:
This PR does the following:

* Uses the `simple-evals` prompt template for math evals to align with what DeepSeek used in their report.
* Uses the same sampling parameters as DeepSeek (`temperature = 0.6` and `top_p = 0.95`)

The result is closer alignment with the scores reported by DeepSeek, modulo the fact we do not yet compute `pass@1` for AIME24, MATH-500, and GPQA. I will address this in a follow-up PR as it requires changes on the `lighteval` side. 

More importantly, I found that without sampling, the DeepSeek distilled models would produce long CoTs that never terminated with a `</think>` tag and thus performance was heavily penalised. The reason we did not encounter this previously is because `lighteval` set the `temperature = 1` for all evals, including those we thought were greedy.

This means you can expect some variation across eval runs until we includeproper `pass@1` estimation. Evals are currently running and I will update them shortly.

## TODO

- [x] Update eval tables

### DIFF
--- a/src/open_r1/evaluate.py
+++ b/src/open_r1/evaluate.py
@@ -28,12 +28,18 @@ from lighteval.utils.language import Language
 
 
 # Prompt template from simple-evals: https://github.com/openai/simple-evals/blob/6e84f4e2aed6b60f6a0c7b8f06bbbf4bfde72e58/math_eval.py#L17
+# MATH_QUERY_TEMPLATE = """
+# Solve the following math problem step by step. The last line of your response should be of the form Answer: $ANSWER (without quotes) where $ANSWER is the answer to the problem.
+
+# {Question}
+
+# Remember to put your answer on its own line after "Answer:", and you do not need to use a \\boxed command.
+# """.strip()
+
 MATH_QUERY_TEMPLATE = """
-Solve the following math problem step by step. The last line of your response should be of the form Answer: $ANSWER (without quotes) where $ANSWER is the answer to the problem.
+Solve the following math problem efficiently and clearly.  The last line of your response should be of the following format: 'Therefore, the final answer is: $\\boxed{{ANSWER}}$. I hope it is correct' (without quotes) where ANSWER is just the final number or expression that solves the problem. Think step by step before answering.
 
 {Question}
-
-Remember to put your answer on its own line after "Answer:", and you do not need to use a \\boxed command.
 """.strip()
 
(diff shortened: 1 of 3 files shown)


src/open_r1/evaluate.py
```diff
@@ -27,6 +27,27 @@
 from lighteval.utils.language import Language
 
 
+# Prompt template from simple-evals: https://github.com/openai/simple-evals/blob/6e84f4e2aed6b60f6a0c7b8f06bbbf4bfde72e58/math_eval.py#L17
+MATH_QUERY_TEMPLATE = """
```
lewtun, line 31: This is a good template as it makes less strong assumptions about whether the model has been trained / prompted to emit boxed outputs

src/open_r1/evaluate.py
```diff
@@ -27,6 +27,27 @@
...
+Solve the following math problem step by step. The last line of your response should be of the form Answer: $ANSWER (without quotes) where $ANSWER is the answer to the problem.
+
+{Question}
+
+Remember to put your answer on its own line after "Answer:", and you do not need to use a \\boxed command.
+""".strip()
+
+# Prompt template from simple-evals: https://github.com/openai/simple-evals/blob/83ed7640a7d9cd26849bcb3340125002ef14abbe/common.py#L14
+GPQA_QUERY_TEMPLATE = """
```
lewtun, line 40: This is the same as before, but I've just moved it up here for better code readability

edbeeching (approved):

src/open_r1/evaluate.py
```diff
@@ -27,6 +27,28 @@
 from lighteval.utils.language import Language
 
 
+# Prompt template adapted from
+# - simple-evals: https://github.com/openai/simple-evals/blob/6e84f4e2aed6b60f6a0c7b8f06bbbf4bfde72e58/math_eval.py#L17
+# - Llama 3: https://huggingface.co/datasets/meta-llama/Llama-3.2-1B-Instruct-evals/viewer/Llama-3.2-1B-Instruct-evals__math__details?views%5B%5D=llama_32_1b_instruct_evals__math__details
+# Note that it is important to have the final answer in a box for math-verify to work correctly
+MATH_QUERY_TEMPLATE = """
```
lewtun, line 34: This template guarantees the model's output the answer in a format compatible with `math-verify`

PR #397, opened 2025-02-23
Update sft.py

zhangshan-zs94:
In the experiment, I found that the label for the instruction part in the SFT process was not set to -100, and the packing operation might introduce some errors, so this parameter needs to be removed.


lewtun:
Hello @zhangshan-zs94 thanks for the PR! Unfortunately using this data collator is too model-specific because one needs to hard-code the response template per chat template, so we will keep the code as-is. You can also disable packing by providing it as a command line argument with `--packing false`

PR #398, opened 2025-02-23
feat: make reward functions to support parallel computation

0x404:
Motivation:
Current reward functions generally take a completion list and calculate rewards for each example in this list serially. In some cases, calculating a reward for a single example may take much time, for example, when running code evaluation in e2b. In such situations, if the reward scores for each example could be calculated in parallel, it would improve our training speed.

Approach:
This PR abstracts out a `BaseRewardFunction` class, where each reward function inherits from this class and implements its own`reward_on_single_completion`. This function will receive the completion for each example to be evaluated, along with its corresponding kwargs.

This refactor should make the code clearer and help us support more reward functions in the future.

### DIFF
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -39,6 +39,19 @@ class BaseRewardFunction(ABC):
         """
         raise NotImplementedError("reward_on_single_completion should be impl by subclass")
 
+    def _single_thread_call(self, completions: List[Dict[str, str]], **kwargs) -> List[float]:
+        results = []
+        for idx, completion in enumerate(completions):
+            # prepare per-completion kwargs
+            per_completion_kwargs = {}
+            for key, value in kwargs.items():
+                if isinstance(value, list):
+                    per_completion_kwargs[key] = value[idx]
+                else:
+                    per_completion_kwargs[key] = value
+            results.append(self.reward_on_single_completion(completion, **per_completion_kwargs))
+        return results
+
     def __call__(self, completions: List[Dict[str, str]], **kwargs) -> List[float]:
         """Process and score multiple model completions in parallel.
 
@@ -61,6 +74,9 @@ class BaseRewardFunction(ABC):
             max_workers = kwargs["max_workers"]
         else:
             max_workers = self.max_workers
+
+        if max_workers == 1:
+            return self._single_thread_call(completions, **kwargs)
 
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
             future_to_idx = {}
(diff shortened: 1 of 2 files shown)


src/open_r1/rewards.py
```diff
@@ -18,20 +20,112 @@
...
+        for idx, completion in enumerate(completions):
+            # prepare per-completion kwargs
+            per_completion_kwargs = {}
+            for key, value in kwargs.items():
+                if isinstance(value, list):
+                    per_completion_kwargs[key] = value[idx]
+                else:
+                    per_completion_kwargs[key] = value
+            results.append(self.reward_on_single_completion(completion, **per_completion_kwargs))
+        return results
```
0x404, line 53: The reason for implementing `_single_thread_call` is that some reward evaluations are not thread-safe. When max_workers==1, we use `_single_thread_call` to sequentially compute the reward for each example.

Currently, `math_verify.parse` is not thread-safe: https://github.com/huggingface/Math-Verify/issues/22

This sequential execution mode ensures correct reward computation for non-thread-safe evaluators, even though it doesn't take advantage of parallel processing capabilities.

0x404:
I'll update to multiprocessing instead of multithreading if you find this PR acceptable. (https://github.com/huggingface/Math-Verify/issues/22#issuecomment-2678088287)

edbeeching (changes requested):
Thanks for adding this, can you add a test that compares single vs multi-thread/proc to ensure they both return the same results.

0x404:
Sure

edbeeching:
Hey @0x404 I am looking at GRPO training with code rewards today, so I will try and get this PR cleanedup and merged.

0x404:
Sure, thanks. I've been busy with other things these couple of days and haven't had time to update this PR.

lewtun:
Hi @0x404 thanks for the PR! Aside from code execution with E2B, do you know which (if any) of the reward functions are slow to execute? 

If it's just the E2B sandbox, I'm wondering if it's better to use their native [async sandbox](https://e2b.dev/docs/sdk-reference/python-sdk/v1.1.0/sandbox_async#async-sandbox) instead of enforcing multi-threading on all rewards.

0x404:
hi, @lewtun, I think currently only e2b rewards are relatively slow

lewtun:
> hi, @lewtun, I think currently only e2b rewards are relatively slow

Thanks, let me take a stab at making it async first since I am quite partial to the simplicity of having simple functions per reward.

edbeeching:
Resolved with aysnc by #484  ?

0x404:
yes

PR #401, opened 2025-02-23
Fix typo in vLLM MODEL_ARGS

Maxwell-Jia:
Fixes a typo in the vLLM parameter name from `gpu_memory_utilization` (American spelling) to `gpu_memory_utilisation` (British spelling) to match vLLM's expected parameter name.


lewtun:
Hi @Maxwell-Jia the British spelling was originally a typo in `lighteval` which was fixed here: https://github.com/huggingface/lighteval/pull/553

As a result, we need to retain the American spelling to compatibility

PR #415, opened 2025-02-24
add sft recipe

eliebak:
Add sft receipe for https://huggingface.co/open-r1/OpenR1-Qwen-7B


lewtun (approved):
Thanks for the recipe! LGTM

PR #418, opened 2025-02-24
Align format reward with R1 traces and add reward function to count think / answer tags

lewtun:
This PR aligns the format reward function to be closer to the traces from DeepSeek-R1, which have the following format:

```html
<think>
{thoughts}
</think>
{response}
```

To remain consistent with the `<answer>` and `</answer>` tags in the existing `format_reward()` function, the new format is now:

```html
<think>
{thoughts}
</think>
<answer>
{response}
</answer>
```

Our prior regex was softer since it didn't enforce line breaks, a detail that is somewhat important if you want your inference code to be consistent across models. Note that the tests had to be updated to accommodate this change.

I also noticed in some GRPO runs that the models can end up emitting multiple thought / answer blocks in a single completion which gives positive reward for the format, but is not ideal. To handle that, I've introduced a tag count function that is inspired from Will's GRPO demo: https://gist.github.com/willccbb/4676755236bb08cab5f4e54a0475d6fb

### DIFF
--- a/src/open_r1/grpo.py
+++ b/src/open_r1/grpo.py
@@ -70,7 +70,7 @@ class GRPOScriptArguments(ScriptArguments):
     reward_funcs: list[str] = field(
         default_factory=lambda: ["accuracy", "format", "tag_count"],
         metadata={
-            "help": "List of reward functions. Possible values: 'accuracy', 'format', 'format_deepseek', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'code', 'code_format'"
+            "help": "List of reward functions. Possible values: 'accuracy', 'format', 'format_deepseek', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', tag_count', 'code', 'code_format'"
         },
     )
     cosine_min_value_wrong: float = field(
--- a/src/open_r1/grpo.py
+++ b/src/open_r1/grpo.py
@@ -34,6 +34,7 @@ from open_r1.rewards import (
     get_repetition_penalty_reward,
     len_reward,
     reasoning_steps_reward,
+    tag_count_reward,
 )
 from open_r1.utils import get_tokenizer
 from open_r1.utils.callbacks import get_callbacks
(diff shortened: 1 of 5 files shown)


src/open_r1/rewards.py
```diff
@@ -62,12 +62,34 @@ def accuracy_reward(completions, solution, **kwargs):
 
 def format_reward(completions, **kwargs):
     """Reward function that checks if the reasoning process is enclosed within <think> and </think> tags, while the final answer is enclosed within <answer> and </answer> tags."""
-    pattern = r"^<think>.*?</think>\s*<answer>.*?</answer>$"
+    pattern = r"^<think>\n.*?\n</think>\n<answer>\n.*?\n</answer>$"
```
lewtun, line 65: Note that this is also aligned with the system prompt we use for the instruct models

src/open_r1/grpo.py
```diff
@@ -67,7 +68,7 @@ class GRPOScriptArguments(ScriptArguments):
     """
 
     reward_funcs: list[str] = field(
-        default_factory=lambda: ["accuracy", "format"],
+        default_factory=lambda: ["accuracy", "format", "tag_count"],
         metadata={
             "help": "List of reward functions. Possible values: 'accuracy', 'format', 'format_deepseek', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'code', 'code_format'"
```
qgallouedec, line 73:
```suggestion
            "help": "List of reward functions. Possible values: 'accuracy', 'format', 'format_deepseek', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', tag_count', 'code', 'code_format'"
```

qgallouedec (approved):
`<think>`
This PR is about adding a count reward, and also align with the the format where you've \n before and after tags. Overall everything look good, and the PR makes sense. I should approve it with some friendly message.
`</think>`
`<answer>`
This makes sense! Looks good to me 😊
`</answer>`

PR #419, opened 2025-02-24
updates max_seq_length to max length due to a bug in trl

edbeeching:
There was a bug in TRL which caused seq_length to default to 1024 even when `max_seq_length` (soft deprecated) was set to another value.

This PR updates our configs in case anyone is running experiments with the bugged version.

Fixed in TRL in PR  https://github.com/huggingface/trl/pull/2947


qgallouedec:
related https://github.com/huggingface/trl/pull/2895

PR #422, opened 2025-02-25
Flash Attention Padding Fix

Kidand:
When I attempted to use the Qwen2.5 7B model for grpo training, I encountered an error: 
```bash
ValueError: You are attempting to perform batched generation with padding_side='right' this may lead to unexpected behaviour for Flash Attention version of Qwen2.
```
This issue can be resolved by replacing 'right' with 'left'.

### DIFF
--- a/src/open_r1/utils/model_utils.py
+++ b/src/open_r1/utils/model_utils.py
@@ -16,6 +16,7 @@ def get_tokenizer(
         model_args.model_name_or_path,
         revision=model_args.model_revision,
         trust_remote_code=model_args.trust_remote_code,
+        padding_side='left',
     )
 
     if training_args.chat_template is not None:


lewtun:
Hello @Kidand this is a bit odd as I've trained the 7B model before 🤔. Can you please share the command you're running?

Kidand:
> Hello @Kidand this is a bit odd as I've trained the 7B model before 🤔. Can you please share the command you're running?

The command used is the one in the README, and the error occurred during the verification process. The command is as follows:

```bash
ACCELERATE_LOG_LEVEL=info accelerate launch --config_file recipes/accelerate_configs/zero2.yaml \
    --num_processes=7 src/open_r1/grpo.py \
    --config recipes/Qwen2.5-Math-7B/grpo/config_simple_rl.yaml
```

cbokpark:
Hello @Kidand, I have the same issue. How did you resolve it? Could you share your solution? What’s the reason for this error?

Issue #427, opened 2025-02-25
TypeError: Can only compare inequalities with Expr


NileZhou:
When I run : (like in readme)
ACCELERATE_LOG_LEVEL=info \
accelerate launch --debug \
    --config_file recipes/accelerate_configs/zero2.yaml \
    --num_processes=7 src/open_r1/grpo.py \
    --config recipes/Qwen2.5-1.5B-Instruct/grpo/config_demo.yaml 

I got this after some steps:
{'loss': 0.0005, 'grad_norm': 0.10229784995317459, 'learning_rate': 1.91044776119403e-06, 'rewards/accuracy_reward': 0.125000003259629, 'rewards/format_reward': 0.1629464402794838, 'reward': 0.2879464402794838, 'reward_std': 0.3753892332315445, 'completion_length': 611.8995819091797, 'kl': 0.012603759765625, 'epoch': 0.01}
  1%|▉                                                                                                | 32/3347 [30:01<49:59:34, 54.29s/it][rank5]: Traceback (most recent call last):
[rank5]:   File "/njfs/train-nlp/zhouyi9/zz/open-r1/src/open_r1/grpo.py", line 273, in <module>
[rank5]:     main(script_args, training_args, model_args)
[rank5]:   File "/njfs/train-nlp/zhouyi9/zz/open-r1/src/open_r1/grpo.py", line 232, in main
[rank5]:     train_result = trainer.train(resume_from_checkpoint=checkpoint)
[rank5]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank5]:   File "/usr/local/lib/python3.11/dist-packages/transformers/trainer.py", line 2171, in train
[rank5]:     return inner_training_loop(
[rank5]:            ^^^^^^^^^^^^^^^^^^^^
[rank5]:   File "/usr/local/lib/python3.11/dist-packages/transformers/trainer.py", line 2531, in _inner_training_loop
[rank5]:     tr_loss_step = self.training_step(model, inputs, num_items_in_batch)
[rank5]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank5]:   File "/usr/local/lib/python3.11/dist-packages/transformers/trainer.py", line 3669, in training_step
[rank5]:     inputs = self._prepare_inputs(inputs)
[...]

PR #429, opened 2025-02-25
Fix verify err. issue 427

NileZhou:
I fixed issue: [#427](https://github.com/huggingface/open-r1/issues/427)
This commit resolves an edge case where dirty data in the training set or model-generated data could lead to math_verify parsing errors, halting the entire training pipeline.


NileZhou:
too much redundent commit, sorry, I will clear my commits and create pull request again

PR #430, opened 2025-02-25
fix reward verify err

NileZhou:
I fixed issue: [#427](https://github.com/huggingface/open-r1/issues/427)
This commit resolves an edge case where dirty data in the training set or model-generated data could lead to math_verify parsing errors, halting the entire training pipeline.


lewtun (approved):
Thanks for the error handling @NileZhou ! LGTM

PR #449, opened 2025-02-28
Simplify dependencies set up for development mode: update makefile and readme

### LINKED ISSUE #448: Simplify deps install for development mode
Some hardware (eg.. my Mac M1) doesn't like certain experimental dep versions, so I have to use CodeSpaces for development. 
Simplifying the setup from a blank machine would make for an easier onboarding.

I noticed you currently don't need flash-attn for running the unit tests, so I suggest the following:

* Add a `make install` entry to the Makefile, with all the steps mentioned in the readme except the one for installing `flash-attn`

ocramz:
* add `make install`
* add README instructions

Closes #448

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,14 @@ export PYTHONPATH = src
 
 check_dirs := src tests
 
+
+# dev dependencies
+install:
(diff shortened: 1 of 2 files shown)


lewtun (approved):
Thanks for the improvement @ocramz !

PR #463, opened 2025-03-03
Extend max_model_length to prevent context truncation

eldarkurtic:
With the existing example command for evaluation, both `max_model_length` and `max_new_tokens` are set to the same value of `32768`. This produces the following warning:
```bash
[ WARNING]: context_size + max_new_tokens=33238 which is greater than self.max_length=32768. Truncating context to 0 tokens. (vllm_model.py:276)
```

We need to increase the `max_model_length` to a larger value to prevent truncating the context. I ran some tests and the proposed `38768` seems to be large enough for all three tasks I've tested so far: AIME, MATH-500, and GPQA-Diamond.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ We use `lighteval` to evaluate models, with custom tasks defined in `src/open_r1
 
 ```shell
 MODEL=deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B
-MODEL_ARGS="pretrained=$MODEL,dtype=bfloat16,max_model_length=32768,gpu_memory_utilization=0.8,generation_parameters={max_new_tokens:32768,temperature:0.6,top_p:0.95}"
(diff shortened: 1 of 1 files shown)


lewtun (commented):
Thanks for the fix @eldarkurtic! Do you see much variation in the evaluation scores once this is included? I believe some of the Qwen models were trained with 32k context, so once we exceed this the model may start producing gibberish.

eldarkurtic:
I've never done a full eval to compare the two because I stopped using 32k as soon as I saw the warning that prompt has been truncated to 0 tokens
```
[ WARNING]: context_size + max_new_tokens=33238 which is greater than self.max_length=32768. Truncating context to 0 tokens. (vllm_model.py:276)
```
But definitely a good point for Qwen models. I assume this could be up to user to decide based on their chosen model for evals.

PR #480, opened 2025-03-05
Remove unimplemented 'format_deepseek' from reward_funcs documentation

A-transformer:
Removed 'format_deepseek' from GRPOScriptArguments help string as it is not implemented in REWARD_FUNCS_REGISTRY and format_reward already covers DeepSeek-style formatting needs.


lewtun (approved):
Thanks for the fix!

PR #484, opened 2025-03-06
Run e2b async sandbox by default

plaguss:
## Description

Refactor of `src/open_r1/rewards.py` `code_reward` function to allow running concurrent test cases. The original version can be recovered by passing `code_reward(completions, run_sync=True)` (set to False by default).

### DIFF
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -14,7 +14,7 @@ from .utils import is_e2b_available
 
 if is_e2b_available():
     from dotenv import load_dotenv
-    from e2b_code_interpreter import AsyncSandbox, Sandbox
+    from e2b_code_interpreter import AsyncSandbox
 
     load_dotenv()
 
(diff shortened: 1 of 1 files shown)


lewtun (commented):
Thanks a lot for adding the async sandbox @plaguss ! Do you think there is much advantage to retain the original synchronous version? If not, it would be fine to remove it and just retain the async one to keep the code a bit simpler. 

Also, were you able to run the training successfully with the config defined here: https://github.com/huggingface/open-r1?tab=readme-ov-file#-training-with-a-code-interpreter ?

plaguss:
> Thanks a lot for adding the async sandbox @plaguss ! Do you think there is much advantage to retain the original synchronous version? If not, it would be fine to remove it and just retain the async one to keep the code a bit simpler.

Yeah there's no need to keep both, it works just the same, refactored now.

> Also, were you able to run the training successfully with the config defined here: https://github.com/huggingface/open-r1?tab=readme-ov-file#-training-with-a-code-interpreter ?

Yes, I tested it and works just as the original sync version

lewtun (approved):
Thanks for iterating! LGTM

PR #493, opened 2025-03-09
FIX: Three Bugs in async E2B code sandbox

rasdani:
The previous implementation and #484 on async E2B sandbox have three bugs.

First in this line, `verification_info` is actually a list. See the list comprehension right above.
https://github.com/huggingface/open-r1/blob/9890a8d9921ecf27784a18896f3b974b357df903/src/open_r1/rewards.py#L373

Here also screenshot from pdb:
Screenshot 2025-03-09 at 00 21 23

Fixed by collecting `languages` first and from then on passing `languages: list[str]` around.

Second bug is `language` was not passed to `run_script`.
https://github.com/huggingface/open-r1/blob/9890a8d9921ecf27784a18896f3b974b357df903/src/open_r1/rewards.py#L418
Even though it expects `language`.
https://github.com/huggingface/open-r1/blob/9890a8d9921ecf27784a18896f3b974b357df903/src/open_r1/rewards.py#L430

Third bug: see comment below.

I also added more specific error handling, as exceptions of these three bugs were handle too generally.

### DIFF
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -428,7 +428,7 @@ async def run_async(scripts: list[str], languages: list[str]) -> list[float]:
 
 async def run_script(sbx, script: str, language: str) -> float:
     try:
-        execution = await sbx.run_code(script, language=language)
+        execution = await sbx.run_code(script, language=language, timeout=30)
     except Exception as e:
         print(f"Error from E2B executor: {e}")
         return 0.0
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -369,13 +369,8 @@ def code_reward(completions, **kwargs) -> list[float]:
         evaluation_script_template.format(code=json.dumps(code), test_cases=json.dumps(json.dumps(info["test_cases"])))
         for code, info in zip(code_snippets, verification_info)
     ]
-    try:
-        rewards = run_async_from_sync(scripts, verification_info["language"])
-
-    except Exception as e:
-        print(f"Error from E2B executor: {e}")
-        rewards = [0.0] * len(completions)
-
+    languages = [info["language"] for info in verification_info]
+    rewards = run_async_from_sync(scripts, languages)
     return rewards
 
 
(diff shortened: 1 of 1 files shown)


rasdani:
Code rewards after fix. Previously got none.
image

rasdani:
A lot of GPU idling, though?
image
image

Training ETA with buggy version, which essentially just learned format was 1.5 hours.

rasdani:
With previous sync version checkout at ``, also  a lot of idling. Had to fix first bug here, too.
image

rasdani:
will look into reusing sandbox across generations...

rasdani:
Even though we set a sandbox timeout of 30, `sbx.run_script()` still runs into it's default timeout of 300s.
Screenshot 2025-03-09 at 14 22 27

Here's how I timed it:
```python
async def run_script(sbx, script: str, language: str) -> float:
    try:
        start_time = time.time()
        execution = await sbx.run_code(script, language=language,)
        end_time = time.time()
        print(f"Script execution time: {end_time - start_time} seconds")
    except e2b.TimeoutException as e:
        end_time = time.time()
        print(f"Script execution time: {end_time - start_time} seconds")
        print(f"TimeoutException from E2B executor: {e}")
        is_running = await sbx.is_running()
        print(f"{is_running=}")
        breakpoint()
        is_running = await sbx.is_running()
        print(f"{is_running=}")
        return 0.0
    except Exception as e:
        print(f"Error from E2B executor: {e}")
        return 0.0
    try:
        return float(execution.text)
    except (TypeError, ValueError):
        return 0.0
```

rasdani:
On a more general note:

I trained with [verifiable-coding-problems-python_decontaminated](https://huggingface.co/datasets/open-r1/verifiable-coding-problems-python_decontaminated). It has a long tail of examples with a lot of test cases.

Even though `evaluate_code()` limits time for subprocesses to 5s, I still created a filtered dataset with examples containing at most six test cases.
https://huggingface.co/datasets/rasdani/verifiable-coding-problems-python_decontaminated_six_test_cases


Since script execution is the bottleneck, one can think about checking code with `ruff` first before executing it.
Unfortunately Python does not have a compiler like Rust, otherwise something [like this](https://www.oxen.ai/blog/training-a-rust-1-5b-coder-lm-with-reinforcement-learning-grpo) would be feasible.

When I ran in to execution timeouts, I also noticed this was due to long code slop. Maybe discarding unreasonably long completions could also help.

edbeeching:
Thanks for the PR @rasdani. I believe most of these issues were fixed in #504 , so I will close this one.

Issue #492, opened 2025-03-08
SFT model make repetitions during the inference phase


0205090923:
SFT according to [OpenR1-Qwen-7B](https://github.com/huggingface/open-r1/tree/main/recipes/OpenR1-Qwen-7B) and the loss during the training phase decreased normally.
But during inference, I find that SFT model produces a lot of repetition until reach the max_generation limit? Any suggestions?

<img width="756" alt="Image" src="https://github.com/user-attachments/assets/6b0ad056-97ae-437c-b341-486a639caa82" />

0205090923:
[OpenR1-Qwen-7B](https://github.com/huggingface/open-r1/tree/main/recipes/OpenR1-Qwen-7B) is this config leading to  OpenR1-Qwen-7B?  But the trained model seems to be much worse, do I have any misunderstandings?

digitalSquirrel1:
Seems like this is caused by incorrect label of eos_token, which means the model cannot learn to stop during SFT training. Specifically, source code in sft.py is `tokenizer.pad_token = tokenizer.eos_token`, and pad_token cannot contribute to loss in the default DataCollatorForLanguageModeling.
Maybe you would like to check this pr. One-line change will work just fine: https://github.com/huggingface/open-r1/pull/494

HwangYej1:
hi,have you reproduced SFT results on math-500? i have the bad results. in training phase loss decrease normally but test results on math-500 is even lower than model which before SFT

0205090923:
After fixing the bugs, I got the desired results

juliancodaforno:
Which fixes do you refer to? Just the 
`tokenizer.pad_token = tokenizer.eos_token` to
```
if tokenizer.pad_token is None:
    tokenizer.pad_token = tokenizer.eos_token
```
or what else did you change, because just changing this line does not help me.

PR #494, opened 2025-03-09
sft learn to generate eos token

digitalSquirrel1:
Avoid 'eos_token' from being labeled as -100 and stopping contribute to loss. 
one-line changing in sft.py. 
correlated issue: https://github.com/huggingface/open-r1/issues/492

### DIFF
--- a/src/open_r1/sft.py
+++ b/src/open_r1/sft.py
@@ -106,7 +106,8 @@ def main(script_args, training_args, model_args):
     # Load tokenizer
     ################
     tokenizer = get_tokenizer(model_args, training_args)
-    tokenizer.pad_token = tokenizer.eos_token
+    if tokenizer.pad_token is None:
+        tokenizer.pad_token = tokenizer.eos_token
 
     ###################
     # Model init kwargs


digitalSquirrel1:
transformers.data.data_collator.DataCollatorForLanguageModeling:

image

qgallouedec (approved):

qgallouedec:
Thanks!

javirandor:
@digitalSquirrel1 What are you using as your new pad token? Qwen has the same token for EOS and PAD. Are you adding a new token in vocab or am I missing sth?

kiseliu:
> transformers.data.data_collator.DataCollatorForLanguageModeling:
> 
> image

So could you reproduce the results, like having a 50%+ performance on AIME24?

Tim-Siu:
I think we normally would replace the `tokenizer_config` of base models with their instruct counterparts?

> @digitalSquirrel1 This works only when starting SFT from `Qwen/Qwen2.5-Math-7B-Instruct`. Other models have the same EOS and PAD token e.g. `Qwen/Qwen2.5-Math-7B` (see [here](https://huggingface.co/Qwen/Qwen2.5-Math-7B/blob/b101308fe89651ea5ce025f25317fea6fc07e96e/tokenizer_config.json#L200-L203)) so this will not work.

I think we normally would replace the `tokenizer_config` of base models with their instruct counterparts?

lewtun:
Thanks for catching this issue with using the data collator and pad tokens @digitalSquirrel1 ! I think a more robust solution would be to expose `pad_token` in `SFTConfig` on the TRL side as this would allow users to configure it per model. For example, for Qwen models, one could then use some a prexisting, but irrelevant token like `"<|fim_pad|>"` as the pad token.

Would you like to open a PR on TRL for this?

lewtun:
Hey everyone, should be fixed by https://github.com/huggingface/open-r1/pull/604

Was a bit of an edge case in how Qwen base models predefine a chat template 🙀

PR #495, opened 2025-03-09
adds support for running GRPO on IOI problems

guipenedo:


guipenedo:
PR should be good, but something in setup.py seems to no longer works and breaks the tests cc @edbeeching

src/open_r1/rewards.py
```diff
@@ -370,7 +371,8 @@ def evaluate_code(code, test_cases):
         for code, info in zip(code_snippets, verification_info)
     ]
     try:
-        rewards = run_async_from_sync(scripts, verification_info["language"])
+        loop = _init_event_loop()
+        rewards = loop.run_until_complete(run_e2b_async(scripts, verification_info["language"]))
```
edbeeching, line 375: I think is it better to just use `asyncio.run(run_e2b_async(scripts, verification_info["language"]))` ? Then we can drop the `loop = _init_event_loop()` line as this is handled by asyncio

guipenedo: asyncio.run is only meant to be called on the top level entry point (typically a main()) see the [docs](https://docs.python.org/3/library/asyncio-task.html)

src/open_r1/rewards.py
```diff
@@ -379,6 +381,42 @@ def evaluate_code(code, test_cases):
...
+        except Exception as e:
+            print(f"Error from Piston worker: {e}")
+            return SubtaskResult()  # score 0.0
+
+    # load problem data. undo separating kwargs by column
+    problems_data = [dict(zip(kwargs.keys(), values)) for values in zip(*kwargs.values())]
+    
+    loop = _init_event_loop()
+    evals = [
+        loop.create_task(run_catch_exceptions(score_subtask(piston_client, problem_data, code, test_batch_size=test_batch_size)))
```
edbeeching, line 412:
```suggestion
        asyncio.create_task(run_catch_exceptions(score_subtask(piston_client, problem_data, code, test_batch_size=test_batch_size)))
```

src/open_r1/rewards.py
```diff
@@ -379,6 +381,42 @@ def evaluate_code(code, test_cases):
...
+        for problem_data, code in zip(problems_data, code_snippets)
+    ]
+    results = loop.run_until_complete(asyncio.gather(*evals))
```
edbeeching, line 415:
```suggestion
    results = asyncio.run_until_complete(asyncio.gather(*evals))
```

edbeeching, line 415:
```suggestion
    results = await asyncio.gather(*evals)
```

src/open_r1/rewards.py
```diff
@@ -379,6 +381,42 @@ def evaluate_code(code, test_cases):
...
+        try:
+            return await task
+        except Exception as e:
+            print(f"Error from Piston worker: {e}")
+            return SubtaskResult()  # score 0.0
+
+    # load problem data. undo separating kwargs by column
+    problems_data = [dict(zip(kwargs.keys(), values)) for values in zip(*kwargs.values())]
+    
+    loop = _init_event_loop()
```
edbeeching, line 410:
```suggestion
```

src/open_r1/rewards.py
guipenedo: yeah this works because we set the asyncio event loop but you do need to keep the _init line otherwise you'll get an error from this

src/open_r1/rewards.py
guipenedo: see other comments

src/open_r1/rewards.py
guipenedo: I can refactor a bit and rename `_init_event_loop` to `get_event_loop` and then just have
```python
rewards = get_event_loop().run_until_complete...
```
if you prefer, but, using `run` or just creating and destroying event loops all the time isn't a good idea

edbeeching (approved):
LGTM barring some asyncio changes.

In #504 I added slow tests, that can be executed locally. Can you add a test that runs the reward function with some C++ code, I have been using gold solutions from the datasets we have been buiding for python reward func slow tests.

src/open_r1/rewards.py
edbeeching: ah sorry, I thought we were still in the async func. Good points, thanks.

src/open_r1/rewards.py
guipenedo: we could test a bit to be sure but in theory each successive call to the reward function should reuse the existing loop (so there will be a single lingering event loop at the very end)

edbeeching: Can you take a look at PR #504 ? I did some fixes / refactor there with asyncio (I don't claim this is the right way to do things!)

I am currently running the code reward on gold answers for the whole `open-r1/verifiable-coding-problems-python_decontaminated` (27k examples) with dataset.map on 16 procs and no issues so far, but I think each proc will have its own loop that is created and destroyed by asyncio.

slurm/piston/launch_piston_workers.sh
```diff
@@ -0,0 +1,16 @@
...
+
+for i in $(seq 1 $N_INSTANCES); do
+    # Find random (hopefully) available port
+    PORT=$(comm -23 <(seq 2000 10000 | sort) <(ss -tan | awk '{print $4}' | cut -d':' -f2 | sort -u) | shuf | head -n1)
+    
+    # the job name format is important for the code to then be able to get a list of workers. `piston-worker-<port>`
+    sbatch \
+        --job-name="piston-worker-$PORT" \
+        --export=ALL,PORT=$PORT \
+        /fsx/guilherme/piston/launch_single_piston.sh
```
min-xu-et, line 15: this hard coded path won't work in the general case?

guipenedo: no, just like other paths in the slurm scripts you will need to adapt them

PR #508, opened 2025-03-15
Fix buffer

lewtun:
Also tunes the Qwen config to work with TRL constraints on num_generations


edbeeching (approved):
thanks, I fixed the buffer I removed locally, but I had not pushed it yet. feel free to merge.

Issue #503, opened 2025-03-12
Fail to parse gold solution


Lynnzake:
- latex2sympy2_extended: 0.7.0
- python: 3.10

```python
def accuracy_reward(completions, solution, **kwargs):
    """Reward function that checks if the completion is the same as the ground truth."""
    ........Exactly the same as the repo.
        else:
            # If the gold solution is not parseable, we reward 1 to skip this example
            reward = 1.0
            print("Failed to parse gold solution: ", sol)
        rewards.append(reward)

    return rewards
```

When I train Qwen-0.5B-instruct with grpo, Using the dataset [NuminaMath-TIR](https://huggingface.co/datasets/AI-MO/NuminaMath-TIR)], but the stdout keeps print `Failed to parse gold solution`. Was the version of my `latex2sympy2_extended` a little update?

Jcorners:
I checked the grpo_trainer in RTL, and for the accuracy reward function, it only passes in the prompt and the answer generated by the model. At the same time, reward_kwargs will pass in other data for the example, such as answer, is_deasoning_complete, source, etc. in the OpenR1-Math-220k dataset. However, the reward function of open-r1 only uses the prompt and answer, so it is not possible to obtain the gold solution. I think it is necessary to rewrite the reward function get gold solution from reward_kwargs.

Lynnzake:
> I checked the grpo_trainer in RTL, and for the accuracy reward function, it only passes in the prompt and the answer generated by the model. At the same time, reward_kwargs will pass in other data for the example, such as answer, is_deasoning_complete, source, etc. in the OpenR1-Math-220k dataset. However, the reward function of open-r1 only uses the prompt and answer, so it is not possible to obtain the gold solution. I think it is necessary to rewrite the reward function get gold solution from reward_kwargs.

I have read the GRPOTrainer today, the reward setting in relevance are below:
```python
keys = [key for key in inputs[0] if key not in ["prompt", "completion"]]
reward_kwargs = {key: [example[key] for example in inputs] for key in keys}
output_reward_func = reward_func(prompts=prompts, completions=completions, **reward_kwargs)
```
It seems like the `solution` item of [NuminaMath-TIR](https://huggingface.co/datasets/AI-MO/NuminaMath-TIR) can be pass to the reward func correctly.
If I change the accuracy_reward func to :
```python
def accuracy_reward(completions, **kwargs):
    # Extract responses
    contents = [completion[0]["content"] for completion in completions]
    rewards = []
[...]

PR #509, opened 2025-03-16
Fix for `Failed to parse gold solution:` error

saidineshpola:
Fix for `Failed to parse gold solution:` error
- GRPO rewards returns only answer (no solution), so remove gold solution by parsing of solution from dataset
- Access the answer directly and compare with math_verify


saidineshpola:
Github reported Issue can be found [here](https://github.com/huggingface/open-r1/issues/503)

Issue #160, opened 2025-02-02
NCCL problem occured when multiple GPU cards are saving model.safetensors


pyh314:
I successfully trained the model Qwen/Qwen2.5-1.5B-Instruct through sft.py. However when I saved the model.safetensors I met the problem below:
```
[rank4]:[E202 22:11:23.884267610 ProcessGroupNCCL.cpp:616] [Rank 4] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=59709, OpType=_ALLGATHER_BASE, NumelIn=29171712, NumelOut=233373696, Timeout(ms)=1800000) ran for 1800018 milliseconds before timing out.
[rank4]:[E202 22:11:23.885041639 ProcessGroupNCCL.cpp:1785] [PG ID 0 PG GUID 0(default_pg) Rank 4] Exception (either an error or timeout) detected by watchdog at work: 59709, last enqueued NCCL work: 59711, last completed NCCL work: 59708.
[rank6]:[E202 22:11:23.891257576 ProcessGroupNCCL.cpp:616] [Rank 6] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=59709, OpType=_ALLGATHER_BASE, NumelIn=29171712, NumelOut=233373696, Timeout(ms)=1800000) ran for 1800026 milliseconds before timing out.
[rank6]:[E202 22:11:23.891933152 ProcessGroupNCCL.cpp:1785] [PG ID 0 PG GUID 0(default_pg) Rank 6] Exception (either an error or timeout) detected by watchdog at work: 59709, last enqueued NCCL work: 59711, last completed NCCL work: 59708.
[rank2]:[E202 22:11:23.894245484 ProcessGroupNCCL.cpp:616] [Rank 2] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=59709, OpType=_ALLGATHER_BASE, NumelIn=29171712, NumelOut=233373696, Timeout(ms)=1800000) ran for 1800028 milliseconds before timing out.
[rank2]:[E202 22:11:23.894987441 ProcessGroupNCCL.cpp:1785] [PG ID 0 PG GUID 0(default_pg) Rank 2] Exception (either an error or timeout) detected by watchdog at work: 59709, last enqueued NCCL work: 59711, last completed NCCL work: 59708.
[...]

PR #513, opened 2025-03-17
add ddp_timeout argument in SFTConfig to avoid timeout error(issue #160)

HowToNameMe:
fix issue #160

### DIFF
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -499,6 +499,9 @@ async def run_script(sbx: AsyncSandbox, script: str, language: str) -> float:
         return float(execution.text)
     except (TypeError, ValueError):
         return 0.0
+    except Exception as e:
+        print(f"Error from E2B executor run_script: {e}")
+        return 0.0
 
 
 def get_reward_funcs(script_args) -> list[Callable]:


edbeeching:
hi @HowToNameMe . Thanks for your PR, I am not sure how just adding this to the config resolves the issue. Do you not need to set a var somewhere? Or does this override a default value?

HowToNameMe:
Thanks for your review. The issue #160 happens when NCCL fails out of the ddp timeout. The parameter corresponding to ddp timeout is in the [transformers](https://github.com/huggingface/transformers/blob/a41e08aa19b43d4b20ed71747b074b28b547562c/src/transformers/training_args.py#L1486) repository and not exposed to the developers of open-r1. To tackle this issue, we need to add the `ddp_timeout` to the SFTConfig and override the default value in transformers(30 minutes) with 24 hours.

edbeeching:
@HowToNameMe , `SFTConfig` is a child class of `TrainingArguments`, so you should be able to set this option at the command line / config file.

HowToNameMe:
@edbeeching Thanks for your guidance. It is possible to override the `ddp_timeout` in command line.

PR #528, opened 2025-03-21
Adds binary code reward

edbeeching:
The [RLEF paper](https://arxiv.org/pdf/2410.02089v1) demonstrates that ~binary rewards can improve performance, here we add a variant of the code reward that is binary and is 1 only when all tests pass.

This PR also refactors the reward registry so it is easier to expand without modifying the GRPO script

### DIFF
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -4,6 +4,7 @@ import asyncio
 import json
 import math
 import re
+from functools import partial, update_wrapper
 from typing import Dict
 
 from latex2sympy2_extended import NormalizationConfig
@@ -367,6 +368,12 @@ def extract_code(completion: str, language: str = "python") -> str:
     return extracted_answer
 
 
+def binary_code_reward(completions, **kwargs) -> list[float]:
+    rewards = code_reward(completions, **kwargs)
+    BINARY_THRESHOLD = 0.99
+    return [1.0 if reward > BINARY_THRESHOLD else 0.0 for reward in rewards]
+
+
 def code_reward(completions, **kwargs) -> list[float]:
     """Reward function that evaluates code snippets using the E2B code interpreter.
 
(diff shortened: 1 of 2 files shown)


src/open_r1/rewards.py
```diff
@@ -367,6 +368,12 @@ def extract_code(completion: str, language: str = "python") -> str:
     return extracted_answer
 
 
+def binary_code_reward(completions, **kwargs) -> list[float]:
+    rewards = code_reward(completions, **kwargs)
+    BINARY_THRESHOLD = 0.99
```
lewtun, line 373: Should this be a function arg that the user can set?

kashif (approved):

src/open_r1/rewards.py
```diff
@@ -367,6 +368,12 @@ def extract_code(completion: str, language: str = "python") -> str:
     return extracted_answer
 
 
+def binary_code_reward(completions, **kwargs) -> list[float]:
```
lewtun, line 371: Should we add a slow unit test for this?

src/open_r1/rewards.py
```diff
@@ -492,3 +499,33 @@ async def run_script(sbx: AsyncSandbox, script: str, language: str) -> float:
         return float(execution.text)
     except (TypeError, ValueError):
         return 0.0
+
+
+def get_reward_funcs(script_args: "GRPOScriptArguments") -> list[Callable]:
```
lewtun, line 504: Would it make sense to unit test this just to sanity check we are getting the expected reward functions?

lewtun (commented):
LGTM with a question on whether to unit test some parts

src/open_r1/rewards.py
edbeeching: good point, I will add this.

src/open_r1/rewards.py
edbeeching: This would require code examples which pass only some of the unit tests. I could probably find these in some of our synthetic datasets, but it would take a while and I need to focus on the async exception refactor.

src/open_r1/rewards.py
edbeeching: I am going to say no for now and we can add the feature if we need to toggle it.

PR #532, opened 2025-03-21
Fixes missing exception in run_script 

edbeeching:
Fixes missing exception in `run_script` causing batch of rewards to be zero which one code eval hangs / throws exeception.

requires #528 to be merged first.


lewtun (approved):
LGTM

PR #535, opened 2025-03-22
fix get_reward_funcs bug

AlanShao-zy:
change the input from `script_args.reward_funcs` to `script_args`


edbeeching (approved):

edbeeching:
Good spot, thanks

PR #546, opened 2025-03-26
Async code reward fixes

edbeeching:
- Workaround for bug with E2B's AsyncSandbox where the sandbox would not through a Timeout Exception when it times out.
- Improves exception handling
- Adds a semophore for parallel code execution

### DIFF
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -1,3 +1,18 @@
+# coding=utf-8
+# Copyright 2025 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 """Reward functions for GRPO training."""
 
 import asyncio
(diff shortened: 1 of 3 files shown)


lewtun:
Added a nice table in https://github.com/huggingface/open-r1/pull/546/commits/d30ce0782557853c332eaa99127c52e02ac80ee4 to be able to better read the E2B benchmarking:

## Benchmark Results

| Sample Size | Parallelization | Execution Time (s) | Mean Reward | Min Reward | Max Reward |
|:-----------:|:---------------:|------------------:|:-----------:|:-----------:|:-----------:|
|     16      |        1        |             32.42 |   0.5089    |   0.0000    |   1.0000    |
|     16      |        4        |              9.79 |   0.5089    |   0.0000    |   1.0000    |
|     16      |       16        |              4.78 |   0.5089    |   0.0000    |   1.0000    |
|     64      |        4        |             39.56 |   0.5239    |   0.0000    |   1.0000    |
|     64      |       16        |             19.86 |   0.5239    |   0.0000    |   1.0000    |
|     64      |       64        |             17.17 |   0.5239    |   0.0000    |   1.0000    |
|     256     |       16        |             74.36 |   0.5681    |   0.0000    |   1.0000    |
|     256     |       64        |             62.47 |   0.5681    |   0.0000    |   1.0000    |
|     256     |       96        |             58.34 |   0.5681    |   0.0000    |   1.0000    |

src/open_r1/rewards.py
```diff
@@ -368,13 +383,13 @@ def extract_code(completion: str, language: str = "python") -> str:
     return extracted_answer
 
 
-def binary_code_reward(completions, **kwargs) -> list[float]:
-    rewards = code_reward(completions, **kwargs)
+def binary_code_reward(completions, num_parallel: int = 2, **kwargs) -> list[float]:
+    rewards = code_reward(completions, num_parallel=num_parallel, **kwargs)
     BINARY_THRESHOLD = 0.99
```
lewtun, line 388: Not related to this PR, but why do we need this threshold? Aren't the rewards always 1 or 0 in this case?

lewtun (approved):
LGTM - note that I've added the sandbox IDs as this is useful information to share with E2B for debugging.

src/open_r1/rewards.py
edbeeching: The rewards are in the range 0-1 depending on how many unit tests pass. I use a threshold as we may want to vary it in the future, but I did not want to expose it yet until we feel the need.

lewtun: Ah I see, in that case wouldn't a threshold of 0.5 be more natural (as one does with e.g. classifiers)? Can naturally be left as something to follow-up on if we're not getting enough signal with the current threshold

PR #533, opened 2025-03-21
[WIP] RL goes brrr

lewtun:
Adds support to run GRPO with `vllm` in multi-node settings. Needs https://github.com/huggingface/trl/pull/3094 to be merged first

## TODO

- [x] Add doc
- [x] Validate we can train multi-node with 7B model
- [x] Pin `trl` to new release

PR #549, opened 2025-03-27
Restore single-node instructions to run GRPO

lewtun:
PR #533 accidentally removed the instructions to run GRPO interactively in single-node settings (which is likely the most common scenario for users). 

This PR restores the instructions and updates them to account for the changes to vLLM serving in TRL. 

cc @edbeeching for viz


edbeeching (approved):
Thanks!

PR #540, opened 2025-03-24
fix dataset parsing error

AlanShao-zy:
support defined question field to fix errors when datasets' question field is not 'problem'

PR #558, opened 2025-03-29
Fixing the default of dataset_prompt_column

toslali-ibm:
A recent update introduced the `dataset_prompt_column` argument to fix errors when a dataset's `question` field is not `problem`. However, this change unintentionally broke the default behavior for most example datasets, which do use the `problem` field. To maintain compatibility, the default value should be set to `problem` instead of `prompt`.

### DIFF
--- a/src/open_r1/configs.py
+++ b/src/open_r1/configs.py
@@ -162,6 +162,6 @@ class GRPOScriptArguments(trl.ScriptArguments):
     )
 
     dataset_prompt_column: str = field(
-        default="prompt",
+        default="problem",
         metadata={"help": "Column to use as prompts for training."},
     )


edbeeching:
Hi @toslali-ibm , thanks for the PR. We changed the default in #540 . What examples are you running? We will update the config.

toslali-ibm:
> Hi @toslali-ibm , thanks for the PR. We changed the default in #540 . What examples are you running? We will update the config.

Hello, thanks for the response. From what I’ve seen, the majority of the GRPO examples (if not all) include datasets with a `problem` column—for example: `open-r1/OpenR1-Math-220k, open-r1/verifiable-coding-problems-python-10k, open-r1/ioi, and DigitalLearningGmbH/MATH-lighteval`. That’s why I thought it might be more straightforward to update the default here rather than adjusting the configs.

lewtun:
Hey @toslali-ibm thanks for catching this! I've opted to update the configs instead in bcause `prompt` is the more common column name in the wild: https://github.com/huggingface/open-r1/pull/574

PR #561, opened 2025-03-30
Adds a E2B router server that executes batches of scripts

edbeeching:
This PR creates a `RoutedSandbox` which is a plug and play replacement for E2B's `Sandbox` class, although it works on batches of scripts. The scripts are routed to an E2B router which executes the scripts in parallel on multiple E2B AsyncSandboxes. Using a semephore to ensure we are not rate limited.

### For GRPO training:
First start the router and get its IP
```shell
sbatch slurm/e2b_router.slurm
```

Then add this line in your config: (for example)
```shell
e2b_router_url: ip-10-53-85-124:8000
```

Only one router is needed for all E2B users on the cluster (once we have the HF PRO account)

Defaults to 20 concurrent sandboxes


src/open_r1/utils/routed_sandbox.py
```diff
@@ -0,0 +1,100 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+
+import requests
+from e2b_code_interpreter.models import Execution, ExecutionError, Result
+
+
+class BatchedRoutedSandbox:
```
aymeric-roucher, line 22: Could the name be simplified to `RoutedSandbox` @edbeeching ?

aymeric-roucher:
@edbeeching the API seems simple and practical enough @edbeeching ! For me only one comment, about possibly simplifying the name.

aymeric-roucher (approved):

slurm/e2b_router.slurm
```diff
@@ -0,0 +1,18 @@
...
+#SBATCH --partition=hopper-cpu
+#SBATCH --mem=16g
+#SBATCH --time=72:00:00
+#SBATCH --cpus-per-task=16
+#SBATCH --output=./logs/%x-%j.out
+#SBATCH --err=./logs/%x-%j.err
+#SBATCH --requeue
+
+echo "Starting job"
+module load cuda/12.4
```
lewtun, line 12: I don't think we need this on the CPU nodes?
```suggestion
```

src/open_r1/rewards.py
```diff
@@ -449,12 +451,30 @@ def evaluate_code(code, test_cases):
...
+            request_timeout=28,
+        )
+
+        rewards = []
+        for execution in executions:
+            try:
+                reward = float(execution.text)
+                rewards.append(reward)
+            except Exception:
+                rewards.append(0.0)  # could this be None?
```
lewtun, line 473: Yes on the TRL side, `None` rewards are excluded from the loss

lewtun (approved):
LGTM with a few nits :)

scripts/e2b_router.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    BatchRequest is a data model representing a batch processing request.
+
+    Attributes:
+        scripts (List[str]): A list of script names or paths to be executed.
+        language (str): The programming language in which the scripts are written.
+        timeout (int): The maximum allowed execution time for each script in seconds.
+        request_timeout (int): The maximum allowed time for the entire batch request in seconds.
+    """
+    scripts: List[str]
```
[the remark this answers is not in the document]
edbeeching: Ye I know, thanks.

PR #567, opened 2025-04-01
soft_overlong_punishment from DAPO paper

shirinyamani:
@qgallouedec @lewtun

### DIFF
--- a/src/open_r1/configs.py
+++ b/src/open_r1/configs.py
@@ -165,16 +165,12 @@ class GRPOScriptArguments(trl.ScriptArguments):
         default="prompt",
         metadata={"help": "Column to use as prompts for training."},
     )
-    
+
     max_completion_len: int = field(
         default=0,
-        metadata={
-            "help": "Maximum length of the completion."
-        },
+        metadata={"help": "Maximum length of the completion."},
     )
     soft_punish_cache: int = field(
         default=0,
-        metadata={
-            "help": "Minimum length of the completion."
-        },
+        metadata={"help": "Minimum length of the completion."},
     )
--- a/src/open_r1/configs.py
+++ b/src/open_r1/configs.py
@@ -92,7 +92,7 @@ class GRPOScriptArguments(trl.ScriptArguments):
 
     Args:
         reward_funcs (`list[str]`):
-            List of reward functions. Possible values: 'accuracy', 'format', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'tag_count', 'code', 'ioi_code', 'code_format'.
+            List of reward functions. Possible values: 'accuracy', 'format', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'tag_count', 'code', 'ioi_code', 'code_format', 'soft_overlong_punishment'
         cosine_min_value_wrong (`float`):
             Minimum reward for cosine scaling for wrong answers.
         cosine_max_value_wrong (`float`):
@@ -105,6 +105,10 @@ class GRPOScriptArguments(trl.ScriptArguments):
             Maximum length for cosine scaling.
         code_language (`str`):
             Language for code format reward.
+        max_completion_len (`int`):
+            Maximum number of characters in completion.
+        soft_punish_cache (`int`):
+            Minimum number of characters in completion.
     """
 
     reward_funcs: list[str] = field(
--- a/src/open_r1/configs.py
+++ b/src/open_r1/configs.py
@@ -92,7 +92,7 @@ class GRPOScriptArguments(trl.ScriptArguments):
 
     Args:
         reward_funcs (`list[str]`):
-            List of reward functions. Possible values: 'accuracy', 'format', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'tag_count', 'code', 'ioi_code', 'code_format', 'soft_overlong_punishment'
+            List of reward functions. Possible values: 'accuracy', 'format', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'tag_count', 'code', 'ioi_code', 'code_format', 'soft_overlong_punishment'.
         cosine_min_value_wrong (`float`):
             Minimum reward for cosine scaling for wrong answers.
         cosine_max_value_wrong (`float`):
--- a/src/open_r1/configs.py
+++ b/src/open_r1/configs.py
@@ -166,8 +166,8 @@ class GRPOScriptArguments(trl.ScriptArguments):
         metadata={"help": "Column to use as prompts for training."},
     )
 
-    max_completion_len: int = field(
-        default=0,
+    max_completion_len: float = field(
+        default=3.6,
         metadata={"help": "Maximum length of the completion."},
     )
     soft_punish_cache: int = field(
--- a/src/open_r1/configs.py
+++ b/src/open_r1/configs.py
@@ -165,3 +165,16 @@ class GRPOScriptArguments(trl.ScriptArguments):
         default="prompt",
         metadata={"help": "Column to use as prompts for training."},
     )
+    
+    max_completion_len: int = field(
+        default=0,
+        metadata={
+            "help": "Maximum length of the completion."
+        },
+    )
+    soft_punish_cache: int = field(
+        default=0,
+        metadata={
+            "help": "Minimum length of the completion."
+        },
+    )
--- a/src/open_r1/configs.py
+++ b/src/open_r1/configs.py
@@ -166,11 +166,11 @@ class GRPOScriptArguments(trl.ScriptArguments):
         metadata={"help": "Column to use as prompts for training."},
     )
 
-    max_completion_len: float = field(
-        default=3.6,
-        metadata={"help": "Maximum length of the completion."},
+    max_completion_len: int = field(
+        default=73728,
+        metadata={"help": "Maximum number of characters in completion."},
     )
     soft_punish_cache: int = field(
-        default=0,
-        metadata={"help": "Minimum length of the completion."},
+        default=14746,
+        metadata={"help": "Minimum number of characters in completion."},
     )
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -540,8 +540,8 @@ def get_soft_overlong_punishment(max_completion_len, soft_punish_cache):
     """
     Reward function that penalizes overlong completions. It is used to penalize overlong completions,
     but not to reward shorter completions. Reference: Eq. (13) from the DAPO paper (https://huggingface.co/papers/2503.14476)
-    Note: In the DAPO paper, completion length is measured in token space, 
-    whereas here it is measured in character space. While these units differ, 
+    Note: In the DAPO paper, completion length is measured in token space,
+    whereas here it is measured in character space. While these units differ,
     they are roughly proportional, meaning this function should yield a similar effect.
     exp: completions = ["hello world", "TRL is fire!"], completion = "TRL is fire!", len(completion)
 
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -539,7 +539,7 @@ async def run_script(script: str, language: str, semaphore: asyncio.Semaphore) -
 def get_soft_overlong_punishment(max_completion_len, soft_punish_cache):
     """
     Reward function that penalizes overlong completions. It is used to penalize overlong completions,
-    but not to reward shorter completions. Reference: DAPO paper:https://huggingface.co/papers/2503.14476
+    but not to reward shorter completions. Reference: Eq. (13) from the DAPO paper (https://huggingface.co/papers/2503.14476)
     Note: In the DAPO paper, completion length is measured in token space, 
     whereas here it is measured in character space. While these units differ, 
     they are roughly proportional, meaning this function should yield a similar effect.
--- a/src/open_r1/rewards.py
+++ b/src/open_r1/rewards.py
@@ -25,9 +25,10 @@ from typing import Callable, Dict
 from latex2sympy2_extended import NormalizationConfig
 from math_verify import LatexExtractionConfig, parse, verify
 
+from .configs import max_completion_len, soft_punish_cache
 from .utils import is_e2b_available
 from .utils.ioi import SubtaskResult, add_includes, get_piston_client_from_env, score_subtask
-from .configs import max_completion_len, soft_punish_cache
+
 
 if is_e2b_available():
     from dotenv import load_dotenv
@@ -535,17 +536,19 @@ async def run_script(script: str, language: str, semaphore: asyncio.Semaphore) -
             except Exception as e:
                 print(f"Error from E2B executor kill with sandbox ID {sandbox.sandbox_id} : {e}")
 
+
 def get_soft_overlong_punishment(max_completion_len, soft_punish_cache):
     """
     Reward function that penalizes overlong completions. It is used to penalize overlong completions,
-    but not to reward shorter completions. Reference: DAPO paper:https://huggingface.co/papers/2503.14476 
+    but not to reward shorter completions. Reference: DAPO paper:https://huggingface.co/papers/2503.14476
     Note: This function is response-level, not token-level.
-    exp: completions = ["hello world", "TRL is fire!"], completion = "TRL is fire!", len(completion) 
-    
+    exp: completions = ["hello world", "TRL is fire!"], completion = "TRL is fire!", len(completion)
+
     Args:
         max_completion_len: Maximum length of the completion
         soft_punish_cache: Minimum length of the completion. If set to 0, no minimum length is applied.
     """
+
     def soft_overlong_punishment_reward(completions, **kwargs):
         """Reward function that penalizes overlong completions."""
         rewards = []
@@ -558,7 +561,9 @@ def get_soft_overlong_punishment(max_completion_len, soft_punish_cache):
             else:
                 rewards.append(-1)
         return rewards
+
     return soft_overlong_punishment_reward
+
 
 def get_reward_funcs(script_args) -> list[Callable]:
     REWARD_FUNCS_REGISTRY = {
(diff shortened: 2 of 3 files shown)


src/open_r1/configs.py
```diff
@@ -165,3 +165,12 @@ class GRPOScriptArguments(trl.ScriptArguments):
         default="prompt",
         metadata={"help": "Column to use as prompts for training."},
     )
+
+    max_completion_len: int = field(
+        default=0,
```
qgallouedec, line 170: What value do they use in the paper? We can approximate the corresponding value in the text space

src/open_r1/rewards.py
```diff
@@ -536,6 +536,34 @@ async def run_script(script: str, language: str, semaphore: asyncio.Semaphore) -
                 print(f"Error from E2B executor kill with sandbox ID {sandbox.sandbox_id} : {e}")
 
 
+def get_soft_overlong_punishment(max_completion_len, soft_punish_cache):
+    """
+    Reward function that penalizes overlong completions. It is used to penalize overlong completions,
+    but not to reward shorter completions. Reference: DAPO paper:https://huggingface.co/papers/2503.14476
+    Note: This function is response-level, not token-level.
```
qgallouedec, line 543: Elaborate a bit on this

```suggestion
    Note: In the DAPO paper, completion length is measured in token space, 
    whereas here it is measured in character space. While these units differ, 
    they are roughly proportional, meaning this function should yield a similar effect.
```

qgallouedec (approved):
Good!
Minor remarks to enhance clarity, but otherwise looks good to me

src/open_r1/configs.py
qgallouedec: This!
How about estimating the average number of characters per token and setting the default values for these args accordingly?

shirinyamani: hmm! we can do that too, but it will not exactly match up the paper!
Plus question: isn't it heavily dependent on the tokenizer ?

qgallouedec: I don't think it is

qgallouedec: I think in the past we had to do this for packing in TRL and we had this value of 3.6. Let's see if you find something similar.

edbeeching: Yes, 3.6 is usually a good approximation, although it does depend somewhat on the tokenizer. More recent models, such as Gemma and Llama have larger vocabularies (256k and 128k respectively) so that might mean the avereage tokens per word is lower. 

If you want to use the actual tokens then we could refactor the reward functions to accept the model config and instantiate a tokenizer?

lewtun: I am in favour of allowing reward function to accept the model's tokenizer and this way we can use `GRPOConfig.max_completion_length` to set the total generation budget and the user doesn't have to do a token -> character conversion in their mind.

If we agree, I think this would be best done as part of the reward function `kwargs` where we could also pass `GRPOTrainer.processing_class`. That way we only instantiate the tokenizer once and not every time the reward function is called.

lewtun: Incidentally, the 3.6 scaling factor is dataset / tokenizer dependent and comes from a certain book 🦜 

Screenshot 2025-04-01 at 15 08 17

qgallouedec: Or, just add `completions_ids` to the arguments passed to the reward functions?

shirinyamani: > Yes, 3.6 is usually a good approximation, although it does depend somewhat on the tokenizer. More recent models, such as Gemma and Llama have larger vocabularies (256k and 128k respectively) so that might mean the average tokens per word is lower.
> 
> If you want to use the actual tokens then we could refactor the reward functions to accept the model config and instantiate a tokenizer?

I agree with this, that is somewhat dependent on the tokenizer; and the avrg per work might be lower for bigger models, 
very simple example; for sentence "TRL is fire!"
gpt-4o tokenizer is ---> `[39749, 382, 6452, 0]` where `TRL` is single token (`39749`)
for deepseek-r1 tokenizer is -----> `[11161, 46, 344, 5902, 3] ` where `TR`,`L` are two token (`11161, 46`)

edbeeching: Perhaps we could just instantiate a class here and use `__call__` ? There are a number of reward functions that could be refactored to be classes as well (in another PR), such as `cosine_scaled_reward` and the code reweards

edbeeching: This would backwards compatible and allow us to instantiate the tokenizer only once.

```
class BaseReward:
    def __init__(script_args, model_config, ... ):
        ...
    def __call__(self, completions, ..., **kwargs):
        ...
```

qgallouedec: One solution, not necessarily an alternative as you can do both, is to pass the tokenized completions to the reward function.

```diff
- output_reward_func = reward_func(prompts=prompts, completions=completions, **reward_kwargs)
+ output_reward_func = reward_func(prompts=prompts, completions=completions, completion_ids=completion_ids, **reward_kwargs)
```

This avoids re-tokenization in this case.

https://github.com/huggingface/trl/blob/78925c34b20c8e5b781e19ef22020fbc13e09af6/trl/trainer/grpo_trainer.py#L822

But yes, let's do all that for a follow-up PR.

tests/test_rewards.py
```diff
@@ -407,6 +408,22 @@ def test_tag_count_rewards_missing_all_tags(self):
         rewards = tag_count_reward(completion)
         self.assertEqual(rewards[0], 0.0)
 
+    def test_soft_overlong_punishment(self):
```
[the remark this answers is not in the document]
shirinyamani: @lewtun I believe Given the current formula, rewards are never positive! 🤔 The function is defined as follows:
	•	If the completion length is less than or equal to L_max - L_cache, the reward is set to 0.
	•	If the completion length is between L_max - Lcache and L_max, the reward is computed as (L_max - L_cache) - |y| / L_cache, Since |y| is larger than L_max - L_cache in this case, the numerator is negative, and so the reward is negative.
	•	If the completion length exceeds L_max, the reward is -1.

In other words, under the current design:
	•	Acceptable completions get a reward of 0.
	•	Overlong completions receive a negative reward (up to -1).

isn't?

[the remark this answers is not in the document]
qgallouedec: I think you're right @shirinyamani
Hence the name "punishment".

But maybe you can add a test for this case:

> If the completion length is between L_max - Lcache and L_max, the reward is computed as (L_max - L_cache) - |y| / L_cache

src/open_r1/configs.py
```diff
@@ -165,3 +165,12 @@ class GRPOScriptArguments(trl.ScriptArguments):
         default="prompt",
         metadata={"help": "Column to use as prompts for training."},
     )
+
+    max_completion_len: float = field(
+        default=3.6,
```
qgallouedec, line 170: I think there's some misunderstanding here: why 3.6? Shouldn't this value be the integer value corresponding to the maximum length beyond which you are penalized?

src/open_r1/rewards.py
```diff
@@ -536,6 +536,36 @@ async def run_script(script: str, language: str, semaphore: asyncio.Semaphore) -
...
+    """
+
+    def soft_overlong_punishment_reward(completions, **kwargs):
+        """Reward function that penalizes overlong completions."""
+        rewards = []
+        for completion in completions:
+            completion_length = len(completion)
+            if completion_length <= max_completion_len - soft_punish_cache:
+                rewards.append(0)
+            elif max_completion_len - soft_punish_cache < completion_length <= max_completion_len:
```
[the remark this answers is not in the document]
qgallouedec: We can still keep it for clarity though

tests/test_rewards.py
```diff
@@ -407,6 +408,29 @@ def test_tag_count_rewards_missing_all_tags(self):
...
+        # 110 > 100, reward should be -1.
+        reward_fn = get_soft_overlong_punishment(max_completion_len=100, soft_punish_cache=20)
+        completions = ["a" * 110]
+        rewards = reward_fn(completions)
+        self.assertEqual(rewards, [-1])
+
+    def test_soft_overlong_punishment_intermediate_completion(self):
+        """Test soft overlong punishment reward function for intermediate length completion."""
+        reward_fn = get_soft_overlong_punishment(max_completion_len=100, soft_punish_cache=20)
+        completions = ["a" * 90]  # 90 is between 80 and 100
```
qgallouedec, line 430: 👍

src/open_r1/configs.py
shirinyamani: to me, it did not make sense, cuz i dont undestand how we can define 3.6 as a len of the completion, but i added it wrt the yesterday conversation! oh wait this it the average token per word? @qgallouedec

src/open_r1/rewards.py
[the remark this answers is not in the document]
shirinyamani: I don't think it's always true tho! e.g., if max_completion_len= 100 and soft_punish_cache = 20, this branch will only run if completion_length is between 81 and 100. If the completion length is less than or equal to 80, the earlier condition (`if completion_length <= max_completion_len - soft_punish_cache`) would apply. If it’s greater than 100, then the final else condition applies.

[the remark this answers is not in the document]
qgallouedec: I meant, just this part: `max_completion_len - soft_punish_cache < completion_length`

src/open_r1/configs.py
qgallouedec: >  oh wait this it the average token per word?
yes, that right

shirinyamani: okay, then should it revert back to 0 ?

src/open_r1/rewards.py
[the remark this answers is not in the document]
shirinyamani: i think so then yes! its true!

src/open_r1/configs.py
qgallouedec: I'd set it to _value used in the DAPO paper_ * 3.6

shirinyamani: so do you mean 16,384 * 3.6 = 58,982.4
does this make sense ?
can you think of any model exceeding this number in generation so that the punishment applies?

qgallouedec: From the paper it should be 20,480 * 3.6 = 73,728, no? and for soft_punish_cache 4096 * 3.6 = 14,746
Indeed, it is quite high. But these are default values that can be modified by the user.

shirinyamani: so a few points;

- that 20,480 is 16,384 + 4,096 so if we plan to scale by 3.6, we need to scale the original 16,384 i think not the sum bc we are seperately scaling the cache  4096 * 3.6 = 14,746,  if we scale 20k by factor of 3.6 and cache by 3.6 this would mean twice scaling which do not reflect the paper. 
- I don't believe such scale (even 16,384 * 3.6 = 58,982.4) would be helpful really, cuz referring back to paper, sounds like they got punishment for the completions exceeding 20k (20,480) but setting the max to roughly 59k will hardly ever do anything to improve the situation but also it might promote very very very long answers, which we don't want (i know these can get ignored by masking when it comes to loss as they are after EOS) but still it's a waste, isn't?
- so I think we either decrease the scale value or maybe not scale at this point,  cuz honestly, I cannot think of any model surpassing 59k even for long CoT, on the other hand, in open-r1 rewards, we have another reward that punish for repeating in the completions, therefore, contradicting what we do here (setting num to high might generate repetitive) setting the default val to so high might hint the (beginner-level) user wrongly i assume!

qgallouedec: Sorry, I reread 1. several times and I don't understand your point. How I see things:
max_len = 20,480 tokens = 73,728 characters
cache = 4,096 tokens = 14,746 characters

- if the generation is shorter than 58,982 character (ie, approx. 16,384 tokens) then 0 reward
- if the generation is between 58,982 and 73,728 characters, (ie, approx 16,384 and 20,480 tokens), then soft punishment,
- if the generation is longer than 73,728 character (ie, approx. 20,480 tokens) then -1 punishment

qgallouedec: > sounds like they got punishment for the completions exceeding 20k (20,480) but setting the max to roughly 59k

yes, but they use the number of tokens, here we use the number of characters, in the end it should be roughly equivalent

qgallouedec: > it might promote very very very long answers

I don't understand how this reward can promote long completions?

shirinyamani: > Sorry, I reread 1. several times and I don't understand your point. How I see things:
> max_len = 20,480 tokens = 73,728 characters
> cache = 4,096 tokens = 14,746 characters

so the 20,480 has the cache 4,096 inside it, right? ( 16,384 + 4,096-- max_com_len + cache)
if we scale 20k ---> 73k then it would mean both (max_com_len + cache) are increasing by this scale, however we are also separately increasing the cache (4,096  ---> 14,746), this would mean twice increase i feel!

shirinyamani: > yes, but they use the number of tokens, here we use the number of characters, in the end it should be roughly equivalent

Ohhh 🙈 I think I make the same mistake that I put a note on for the user to know what level is this completion discussed!  Yes you are right!

qgallouedec: > increasing by this scale

I wouldn't say "increasing", just converted to number of characters.

shirinyamani: Yes, now makes sense! I was thinking of tokens as the paper!

qgallouedec: > we are also separately increasing the cache (4,096 ---> 14,746), this would mean twice increase i feel!

14,746 characters corresponds to roughly 4096 tokens, so I don't think we increase anything here

shirinyamani: Yes, makes sense!

qgallouedec: Maybe it's worth documenting though, to avoid future confusion. You've two options:
1. `max_completion_len` expressed as the number of tokens, is this case your convert to the number of characters in the function
2. `max_completion_len` expressed as the number of characters, in this case, nothing special to do.

But in both case it has to be clearly documented

shirinyamani: > > we have another reward that punish for repeating in the completions, therefore, contradicting what we do here
> 
> Which reward function are you referring to? If we have such reward function, then they would somehow achieve the same goal, i.e. punish long repetitive completions. One based solely on the length, and the other based solely on the repetition, no?

I was referring to [this one](https://github.com/huggingface/open-r1/blob/2636a2130fac92cec79b0781101dedb888a9aa18/src/open_r1/rewards.py#L298).

qgallouedec: Ok now it looks 👍

src/open_r1/configs.py
```diff
@@ -165,3 +165,12 @@ class GRPOScriptArguments(trl.ScriptArguments):
         default="prompt",
         metadata={"help": "Column to use as prompts for training."},
     )
+
+    max_completion_len: int = field(
```
qgallouedec, line 169: Can you also document in line 107?

shirinyamani: done!

src/open_r1/configs.py
```diff
@@ -92,7 +92,7 @@ class GRPOScriptArguments(trl.ScriptArguments):
 
     Args:
         reward_funcs (`list[str]`):
-            List of reward functions. Possible values: 'accuracy', 'format', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'tag_count', 'code', 'ioi_code', 'code_format'.
+            List of reward functions. Possible values: 'accuracy', 'format', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'tag_count', 'code', 'ioi_code', 'code_format', 'soft_overlong_punishment'
```
qgallouedec, line 95:
```suggestion
            List of reward functions. Possible values: 'accuracy', 'format', 'reasoning_steps', 'cosine', 'repetition_penalty', 'length', 'tag_count', 'code', 'ioi_code', 'code_format', 'soft_overlong_punishment'.
```

src/open_r1/configs.py
```diff
@@ -105,6 +105,10 @@ class GRPOScriptArguments(trl.ScriptArguments):
             Maximum length for cosine scaling.
         code_language (`str`):
             Language for code format reward.
+        max_completion_len (`int`):
+            Maximum number of characters in completion.
```
qgallouedec, line 109:
```suggestion
            Maximum number of characters in completion beyond which a -1 reward is given for soft overlong punishment.
```

qgallouedec: Btw it's not very clear from the variable name that it relates to the soft overlong punishment.
When running the script they'll have these two args which are very close but are for completely different purpose: `max_completion_length` and `max_completion_len`

src/open_r1/configs.py
edbeeching: @qgallouedec I think the class refactor I propose above can achieve the correct results, calculating the reward on tokens rather than words, without needing a refactor on the TRL side. Any comment on why we do not do this now?

lewtun:
@shirinyamani feel free to merge whenever you want :)

edbeeching:
@lewtun can we hold off merging until we have some clarity regarding my question above:

> @qgallouedec I think the class refactor I propose above can achieve the correct results, calculating the reward on tokens rather than words, without needing a refactor on the TRL side. Any comment on why we do not do this now?

qgallouedec:
@edbeeching What you're proposing sounds good to me and should work. Wdyt @shirinyamani?

qgallouedec:
Related: https://github.com/huggingface/trl/issues/3222

shirinyamani:
> What you're proposing sounds good to me and should work. Wdyt @shirinyamani?

Agree! We can do Ed's suggestion first and this will be tackled!

qgallouedec:
https://github.com/huggingface/trl/pull/3272

PR #591, opened 2025-04-11
Fix eval system prompt

edbeeching:
When we launch the evals, the command includes `bash -c ...`, it seems that this does not play well with special characters. 

This PR encodes the system prompt in base64, then decodes it in the sbatch script, as a workaround to the issue.


lewtun (approved):
LGTM!

PR #592, opened 2025-04-11
E2B Router bug fixes

edbeeching:
- Fixes a bug in the E2B router when the semaphore was released before the sandbox was killed, which led to rate limiting when there were many calls being made to the router.
- Fixes a bug the the `RoutedSandbox` to catch when the case execution is None, typically this is where the sandbox timesout.

### DIFF
--- a/tests/slow/test_code_reward.py
+++ b/tests/slow/test_code_reward.py
@@ -62,7 +62,9 @@ class TestCodeRewards(unittest.TestCase):
             return examples
 
         code_dataset = code_dataset["train"].select(range(NUM_SAMPLES))
-        code_dataset = code_dataset.map(batch_code_reward, batched=True, batch_size=BATCH_SIZE, num_proc=4)
+        code_dataset = code_dataset.map(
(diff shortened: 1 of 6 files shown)


tests/slow/test_code_reward.py
```diff
@@ -62,7 +62,9 @@ def batch_code_reward(examples):
             return examples
 
         code_dataset = code_dataset["train"].select(range(NUM_SAMPLES))
-        code_dataset = code_dataset.map(batch_code_reward, batched=True, batch_size=BATCH_SIZE, num_proc=4)
+        code_dataset = code_dataset.map(
+            batch_code_reward, batched=True, batch_size=BATCH_SIZE, num_proc=4, load_from_cache_file=False
```
lewtun, line 66: Why is disabling the cache needed?

lewtun (approved):
Nice fixes to subtle bugs! LGTM

tests/slow/test_code_reward.py
edbeeching: If you rerun the test locally, it uses the cached output instead of rerunning it.

PR #600, opened 2025-04-14
Set DP=1 due to vLLM <> LightEval hanging

lewtun:
Workaround for:  https://github.com/huggingface/lighteval/issues/670

Note that I have not updated the main entrypoint in the README as it's better to just fix the root cause than change everything twice


edbeeching (approved):

src/open_r1/utils/evaluation.py
```diff
@@ -73,7 +73,7 @@ def run_lighteval_job(
     if get_param_count_from_repo_id(model_name) >= 30_000_000_000:
```
edbeeching, line 73: Could we lower the threshold so that throughput is higher for 7B models?

PR #611, opened 2025-04-17
Fix evals again for TP>1 

lewtun:


src/open_r1/utils/evaluation.py
```diff
@@ -72,12 +72,10 @@ def run_lighteval_job(
     num_gpus = get_gpu_count_for_vllm(model_name, model_revision)
     if get_param_count_from_repo_id(model_name) >= 30_000_000_000:
         tensor_parallel = True
-    else:
-        num_gpus = 8
-        tensor_parallel = False
-
     # FIXME: vLLM 0.8.3 hangs with lighteval and DP > 1, so we disable it for now and use TP for all evals. See https://github.com/huggingface/lighteval/issues/670
-    tensor_parallel = True
```
lewtun, line 80: This conflicts with Qwen models which require the number of attention heads to be divisible by TP. The fix is to disable DP altogether and rely on https://github.com/huggingface/open-r1/pull/611/files#diff-0118d8beccd2ee9e859eef8c93702e25993678ab87337c7f394e0b622bb0ba2cL72 to get the correct number of GPUs

PR #614, opened 2025-04-17
Code Execution using Morph Cloud

abetlen:
# Code Execution using Morph Cloud

## Overview

This PR introduces [Morph Cloud](https://cloud.morph.so/web/) as an optional backend for code execution within Open-R1, providing users with an alternative provider for Codeforces as well as IOI code reward calculation.

## Key Changes & Design

The integration supports two distinct execution pathways, chosen via configuration:

### 1. General Code Execution (`code_reward`)

**Mechanism**: Integrated via `src/open_r1/utils/code_providers.py::MorphProvider`
This provider supports two operational modes:
- **Router Mode**: If `morph_router_url` is provided in the configuration, the provider uses a client (`src/open_r1/utils/routed_morph.py::RoutedMorphSandbox`) to communicate with a separate router service (`scripts/morph_router.py`). This pattern aims to improve performance and manage costs via sandbox reuse, similar to the E2B router.
- **Direct Mode**: If `morph_router_url` is not provided, `MorphProvider` falls back to interacting directly with the morphcloud SDK (`MorphCloudClient`, `Sandbox`). It manages concurrent sandbox creation and execution using `asyncio` up to the limit specified by `num_parallel`.
**Configuration**: Enabled by setting `provider_type: “morph”`. The `morph_router_url` field is optional; omitting it selects Direct Mode (`src/open_r1/configs.py`). A Slurm script (`slurm/morph_router.slurm`) is provided for launching the router service if Router Mode is chosen.

### 2. IOI Problem Execution (`ioi_code_reward`)
**Mechanism**: Integrated via a custom client implementation (`src/open_r1/utils/ioi/morph_client.py::MorphCloudExecutionClient`). This client uses the lower-level APIs from the morphcloud library (`morphcloud.api.MorphCloudClient`) to manage Morph snapshots and instances directly. It orchestrates the complex setup required for IOI problems (multi-file uploads, custom compile/run scripts) for fine-grained control. This pathway always interacts directly with Morph Cloud and does not use the router.
**Configuration**: Enabled by setting `ioi_provider: “morph”` in the training configuration (`src/open_r1/configs.py`).

## Supporting Changes
- Updated core reward logic in `src/open_r1/rewards.py` to dispatch to the appropriate Morph provider/mode based on configuration settings.
- Added necessary configuration fields to dataclasses in `src/open_r1/configs.py`.
- Included comprehensive setup instructions and usage guidance in `README.md` and a dedicated technical overview in `MORPH_INTEGRATION.md`.
- Updated relevant example configuration files in `recipes/`.

## Dependencies & Usage
- Requires the `morphcloud` package (install via `uv pip install morphcloud`). [Note: Verify addition to project dependencies, e.g., in `setup.py`]
- Users must obtain a Morph API key and set it as the `MORPH_API_KEY` environment variable. This key is used by the Morph router service (if used) and the direct clients (`MorphProvider` in Direct Mode, `MorphCloudExecutionClient` for IOI).
- Activation is opt-in via configuration flags.

## Backward Compatibility
- This change is fully backward compatible.
- E2B remains the default provider for `code_reward` (`provider_type=“e2b”`).
- Piston remains the default provider for `ioi_code_reward` (`ioi_provider=“piston”`).
- Existing workflows using the default providers are unaffected.


README.md
```diff
@@ -227,23 +227,56 @@ You will need to adapt the `slurm/train.slurm` script to match your cluster.
...
 E2B_API_KEY="e2b_xxx"
 ```
 
-Then make sure your dataset contains a `verification_info` column with the following schema (adopted from PrimeIntellect's excellent [datasets](https://huggingface.co/collections/PrimeIntellect/synthetic-1-67a2c399cfdd6c9f7fae0c37) of verifiable problems):
+##### Morph Provider
+
+To use Morph, first install the morphcloud package:
+
+```shell
+pip install morphcloud
```
lewtun, line 254: WDYT about adding this to the `code` extra in `setup.py` so users can just run `uv pip install -e '.[code]'`?

src/open_r1/rewards.py
```diff
@@ -450,43 +472,35 @@ def evaluate_code(code, test_cases):
...
+            test_cases=json.dumps(json.dumps(info["test_cases"]))
+        )
         for code, info in zip(code_snippets, verification_info)
     ]
 
     language = verification_info[0]["language"]
-    if not all(v["language"] == language for v in verification_info):
-        raise ValueError("All verification_info must have the same language", verification_info)
+    
+    if enforce_same_language:
```
lewtun, line 492: Nice check!

src/open_r1/rewards.py
```diff
@@ -398,19 +407,32 @@ def binary_code_reward(completions, num_parallel: int = 2, e2b_router_url=None,
     return output
 
 
-def code_reward(completions, num_parallel: int = 2, e2b_router_url=None, **kwargs) -> list[float]:
-    """Reward function that evaluates code snippets using the E2B code interpreter.
+# This function has been removed as we now use a unified approach in code_reward
```
lewtun, line 410: I think we can remove these comments and `code_reward_morph` since it doesn't seem to be used anywhere?

src/open_r1/rewards.py
```diff
@@ -384,8 +386,15 @@ def extract_code(completion: str, language: str = "python") -> str:
     return extracted_answer
 
 
-def binary_code_reward(completions, num_parallel: int = 2, e2b_router_url=None, **kwargs) -> list[float]:
-    rewards = code_reward(completions, num_parallel=num_parallel, e2b_router_url=e2b_router_url, **kwargs)
+def binary_code_reward(completions, num_parallel: int = 2, e2b_router_url=None, provider_type: str = None, enforce_same_language: bool = False, **kwargs) -> list[float]:
```
lewtun, line 389: I realise this is a breaking change, but maybe it's better to do this than have a router URL arg for E2B and then have users remember they need `kwargs` for Morph:
```suggestion
def binary_code_reward(completions, num_parallel: int = 2, provider_type: str = None, enforce_same_language: bool = False, **kwargs) -> list[float]:
```

The alternative would be to expose

src/open_r1/rewards.py
```diff
@@ -384,8 +386,15 @@ def extract_code(completion: str, language: str = "python") -> str:
...
+    rewards = code_reward(
+        completions, 
+        num_parallel=num_parallel, 
+        e2b_router_url=e2b_router_url,
```
lewtun, line 393: Ditto here if you agree with the above change:
```suggestion
```

src/open_r1/rewards.py
```diff
@@ -398,19 +407,32 @@ def binary_code_reward(completions, num_parallel: int = 2, e2b_router_url=None,
...
+    """This function has been deprecated in favor of the unified code_reward function.
+    
+    All code evaluation now goes through the same code_reward function, regardless of provider.
+    """
+    # TODO: implement support for c++/rust/javascript
+
+    return code_reward(completions, num_parallel=num_parallel, provider_type="morph", **kwargs)
+
+
+def code_reward(completions, num_parallel: int = 2, e2b_router_url=None, provider_type: str = "e2b", enforce_same_language: bool = False, **kwargs) -> list[float]:
```
lewtun, line 422:
```suggestion
def code_reward(completions, num_parallel: int = 2, provider_type: str = "e2b", enforce_same_language: bool = False, **kwargs) -> list[float]:
```

src/open_r1/rewards.py
```diff
@@ -450,43 +472,35 @@ def evaluate_code(code, test_cases):
...
-        rewards = run_async_from_sync(scripts, language, num_parallel)
-    except Exception as e:
-        print(f"Error from E2B executor: {e}")
-        rewards = [0.0] * len(completions)
-
-    return rewards
+    execution_provider = get_provider(
+        provider_type=provider_type,
+        num_parallel=num_parallel,
+        e2b_router_url=e2b_router_url,
```
lewtun, line 500: We should propagate `kwargs` here to use Morph right?
```suggestion
        **kwargs,
```

lewtun (approved):
Thanks a lot for the nice feature @abetlen ! Overall the changes LGTM with some small suggestions to align the function args

scripts/morph_router.py
```diff
@@ -0,0 +1,173 @@
...
+    BatchRequest is a data model representing a batch processing request.
+
+    Attributes:
+        scripts (list[str]): A list of script names or paths to be executed.
+        language (str): The programming language in which the scripts are written.
+        timeout (int): The maximum allowed execution time for each script in seconds.
+        request_timeout (int): The maximum allowed time for the entire batch request in seconds.
+    """
+    scripts: List[str]
+    language: str
```
edbeeching, line 39: One assumption I made when I wrote the E2B router was that all scripts were in the same language, should we relax this assumption here?

slurm/morph_router.slurm
```diff
@@ -0,0 +1,25 @@
...
+#SBATCH --nodes=1
+#SBATCH --ntasks-per-node=1
+#SBATCH --cpus-per-task=8
+#SBATCH --mem=32G
+#SBATCH --time=24:00:00
+#SBATCH --output=logs/morph_router_%j.out
+#SBATCH --error=logs/morph_router_%j.err
+
+# Load necessary modules
+module load anaconda
```
edbeeching, line 12: I understand that you were testing this with you own conda env, but for the PR can you copy paste the `e2b_router.slurm` file and include your `python scripts/morph_router.py --port 8001 --max_num_sandboxes 20`. This will only work on our cluster, but at least the slurm files will be consistent across the project.

src/open_r1/utils/routed_morph.py
```diff
@@ -0,0 +1,117 @@
...
+            router_url: The URL of the MorphCloud router, including host and port.
+            timeout: Default execution timeout in seconds.
+            request_timeout: Default HTTP request timeout in seconds.
+        """
+        self.router_url = router_url
+        self.timeout = timeout
+        self.request_timeout = request_timeout
+    
+    def run_code(self, scripts: List[str], 
+                language: str = "python",
```
edbeeching, line 48: As mentioned above, perhaps we can relax the language here and have a list of languages?

edbeeching (changes requested):
LGTM in general, we have some slow_tests for the code rewards / routers that are not part of the CI in https://github.com/huggingface/open-r1/blob/main/tests/slow/test_code_reward.py

If you could update / add an example there it would be appreciated so we can validate it all works on the cluster.

src/open_r1/utils/routed_sandbox.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, router_url: str):
     def run_code(
         self,
         scripts: list[str],
-        language: str = "python",
+        languages: Optional[List[str]] = None,
```
arb8020, line 44: i made the assumption that i should change the e2b router to support `languages` as well to keep the API consistent between the two but i can also leave the old `language` parameter for backwards compatibility if that's preferred

tests/slow/test_code_reward.py
```diff
@@ -20,12 +20,14 @@
...
 from open_r1.utils.routed_sandbox import RoutedSandbox
+from open_r1.utils.routed_morph import RoutedMorphSandbox
 
 
 class TestCodeRewards(unittest.TestCase):
+    
     def test_python_code_reward(self):
         # requires E2B, see the README.md file
-        code_dataset = load_dataset("open-r1/verifiable-coding-problems-python_decontaminated-tested")
+        code_dataset = load_dataset("open-r1/verifiable-coding-problems-python_decontaminated-tested-shuffled")
```
arb8020, line 30: couldn't find this dataset without the `-shuffled` prefix so i added this

tests/slow/test_code_reward.py
```diff
@@ -90,11 +92,11 @@ def test_e2b_router_run_code_success():
 
         for result in results:
             assert isinstance(result, Execution)
-            assert result.exit_code == 0
+            # assert result.exit_code == 0
             assert result.error is None
-            assert "hello" in result.stdout or "4" in result.stdout
+            assert "hello" in result.logs['stdout'][0] or "4" in result.logs['stdout'][0]
```
arb8020, line 97: also changed a bit of the handling here: maybe something has changed in the Execution object but i wasn't able to access the `.exit_code` attribute, nor `.stdout` without the changes i made here

lewtun (approved):
Thanks for iterating @abetlen ! The slow test LGTM along with the other changes - let's merge once @edbeeching has given it a final pass

edbeeching (approved):
Thanks for the updates, LGTM.

lewtun:
@abetlen can you just fix the merge conflicts and tests and then we're good to go!

lewtun:
Thanks for iterating @abetlen !

PR #629, opened 2025-04-28
Bump LightEval to enable DP>1

lewtun:
This PR bumps `lighteval` to enable DP>1 again since it is compatible with `vllm` v0.8.4. See: https://github.com/huggingface/lighteval/issues/670#issuecomment-2835079312

Waiting for evals to run, but code can be reviewed as is.

Update: evals finished and the diff is up to a few percentage points, but most on AIME24 which is rather noisy due to small sample size:

output-46

Code snippet to run all evals:

```shell
#!/bin/bash

MODELS=(
"DeepSeek-R1-Distill-Qwen-1.5B"
"DeepSeek-R1-Distill-Qwen-7B"
"DeepSeek-R1-Distill-Qwen-14B"
"DeepSeek-R1-Distill-Qwen-32B"
"DeepSeek-R1-Distill-Llama-8B"
"DeepSeek-R1-Distill-Llama-70B"
)

for M in "${MODELS[@]}"; do
  echo "Running benchmark for model: $M"
  python scripts/run_benchmarks.py --model-id deepseek-ai/$M --benchmarks aime24 math_500 gpqa lcb
done
```

## TODO

- [x] Re-run evals to check score variance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ evaluate:
 		fi \
 	),))
 	$(if $(filter tensor,$(PARALLEL)),export VLLM_WORKER_MULTIPROC_METHOD=spawn &&,) \
-	MODEL_ARGS="pretrained=$(MODEL),dtype=bfloat16,$(PARALLEL_ARGS),max_model_length=32768,max_num_batched_tokens=32768,gpu_memory_utilization=0.8,generation_parameters={max_new_tokens:32768,temperature:0.6,top_p:0.95}" && \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -40,7 +40,7 @@ evaluate:
 		fi \
 	),))
 	$(if $(filter tensor,$(PARALLEL)),export VLLM_WORKER_MULTIPROC_METHOD=spawn &&,) \
-	MODEL_ARGS="pretrained=$(MODEL),dtype=bfloat16,$(PARALLEL_ARGS),max_model_length=32768,max_num_batched_tokens=32768,gpu_memory_utilization=0.8,generation_parameters={max_new_tokens:32768,temperature:0.6,top_p:0.95}" && \
```
lewtun, line 43: `max_num_batched_tokens` is no longer needed to be included, so I've removed it for simplicity

edbeeching (approved):

PR #637, opened 2025-05-09
Add dataset filtering script

edbeeching:
This PR adds a script that generates `k` responses from a model and scores the reward using reward function(s).
The mean reward can be used to filter the dataset of easy and hard problems, which may improve GRPO performance.

Usage:

```shell
python scripts/filter_dataset.py --config recipes/dataset_filtering/config_demo.yaml
```

Example dataset, run on a 1k subset:

[OpenR1-Math-220k-generated](https://huggingface.co/datasets/open-r1/OpenR1-Math-220k-generated)

[OpenR1-Math-220k-generated-filtered](https://huggingface.co/datasets/open-r1/OpenR1-Math-220k-generated-filtered)

### DIFF
--- a/recipes/dataset_filtering/config_demo.yaml
+++ b/recipes/dataset_filtering/config_demo.yaml
@@ -1,0 +1,27 @@
+# Model arguments
+model_name_or_path: deepseek-ai/DeepSeek-R1-Distill-Qwen-1.5B
+model_revision: main
+torch_dtype: bfloat16
+attn_implementation: flash_attention_2
+
+# Data training arguments
+# We edit the DeepSeek chat template to ensure (a) the reasoning block within <think> and </think> is included in the completion and (b) the <think> tag is not part of the prefill so that the format reward works
+chat_template: "{% if not add_generation_prompt is defined %}{% set add_generation_prompt = false %}{% endif %}{% set ns = namespace(is_first=false, is_tool=false, is_output_first=true, system_prompt='') %}{%- for message in messages %}{%- if message['role'] == 'system' %}{% set ns.system_prompt = message['content'] %}{%- endif %}{%- endfor %}{{bos_token}}{{ns.system_prompt}}{%- for message in messages %}{%- if message['role'] == 'user' %}{%- set ns.is_tool = false -%}{{'<｜User｜>' + message['content']}}{%- endif %}{%- if message['role'] == 'assistant' and message['content'] is none %}{%- set ns.is_tool = false -%}{%- for tool in message['tool_calls']%}{%- if not ns.is_first %}{{'<｜Assistant｜><｜tool▁calls▁begin｜><｜tool▁call▁begin｜>' + tool['type'] + '<｜tool▁sep｜>' + tool['function']['name'] + '\\n' + '```json' + '\\n' + tool['function']['arguments'] + '\\n' + '```' + '<｜tool▁call▁end｜>'}}{%- set ns.is_first = true -%}{%- else %}{{'\\n' + '<｜tool▁call▁begin｜>' + tool['type'] + '<｜tool▁sep｜>' + tool['function']['name'] + '\\n' + '```json' + '\\n' + tool['function']['arguments'] + '\\n' + '```' + '<｜tool▁call▁end｜>'}}{{'<｜tool▁calls▁end｜><｜end▁of▁sentence｜>'}}{%- endif %}{%- endfor %}{%- endif %}{%- if message['role'] == 'assistant' and message['content'] is not none %}{%- if ns.is_tool %}{{'<｜tool▁outputs▁end｜>' + message['content'] + '<｜end▁of▁sentence｜>'}}{%- set ns.is_tool = false -%}{%- else %}{% set content = message['content'] %}{{'<｜Assistant｜>' + content + '<｜end▁of▁sentence｜>'}}{%- endif %}{%- endif %}{%- if message['role'] == 'tool' %}{%- set ns.is_tool = true -%}{%- if ns.is_output_first %}{{'<｜tool▁outputs▁begin｜><｜tool▁output▁begin｜>' + message['content'] + '<｜tool▁output▁end｜>'}}{%- set ns.is_output_first = false %}{%- else %}{{'\\n<｜tool▁output▁begin｜>' + message['content'] + '<｜tool▁output▁end｜>'}}{%- endif %}{%- endif %}{%- endfor -%}{% if ns.is_tool %}{{'<｜tool▁outputs▁end｜>'}}{% endif %}{% if add_generation_prompt and not ns.is_tool %}{{'<｜Assistant｜>'}}{% endif %}"
+dataset_name: open-r1/OpenR1-Math-220k
+dataset_prompt_column: problem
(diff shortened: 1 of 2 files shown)


recipes/dataset_filtering/config_demo.yaml
```diff
@@ -0,0 +1,27 @@
...
+dataset_name: open-r1/OpenR1-Math-220k
+dataset_prompt_column: problem
+system_prompt: "You are a helpful AI Assistant that provides well-reasoned and detailed responses. You first think about the reasoning process as an internal monologue and then provide the user with the answer. Respond in the following format: <think>\n...\n</think>\n<answer>\n...\n</answer>"
+
+# Generation arguments
+max_completion_length: 2048
+num_generations: 4
```
lewtun, line 16: Let's set 8 as a good default
```suggestion
num_generations: 8
```

scripts/filter_dataset.py
lewtun, line 1: To make it explicit, let's call this `compute_pass_rate.py`

scripts/filter_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+    
+    dataset = dataset.map(apply_chat_template, fn_kwargs={"tokenizer": tokenizer})
+    llm = LLM(
+        model=model_args.model_name_or_path,
+        revision=model_args.model_revision,
+        trust_remote_code=model_args.trust_remote_code,
+    )
+
+    sampling_params=SamplingParams(
+        temperature=training_args.temperature,
```
lewtun, line 103: Expose `top_p` here

scripts/filter_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+            reward_kwargs = {key: repeat_each_element_k_times(examples[key], training_args.num_generations) for key in keys}
+            output_reward_func = reward_func(prompts=repeated_prompts, completions=reward_completions, **reward_kwargs)
+            # Convert None values to NaN
+            output_reward_func = [reward if reward is not None else torch.nan for reward in output_reward_func]
+
+            rewards_per_func[:, i] = torch.tensor(output_reward_func, dtype=torch.float32)
+            
+        reshaped_rewards = rewards_per_func.view(-1, training_args.num_generations)
+        
+        examples["filter_generations"] = grouped_completions
```
lewtun, line 145: Let's be explicit here since filter can mean many things:
```suggestion
        examples["pass_rate_generations"] = grouped_completions
```

scripts/filter_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+        examples["filter_rewards"] = reshaped_rewards.tolist()
```
lewtun, line 146:
```suggestion
        examples["pass_rate_rewards"] = reshaped_rewards.tolist()
```

scripts/filter_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+
```
lewtun, line 147:
```suggestion
            example["pass_rate_generator"] = model_args.model_name_or_path
```

scripts/filter_dataset.py
lewtun: Let's also add a `pass_rate` column, so it's easy to do additional filtering if the original thresholds aren't optimal

scripts/filter_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+    else:
+        unfiltered_dataset_name = f"{script_args.dataset_name}-generated"
+    filtered_dataset_name = f"{unfiltered_dataset_name}-filtered"
+    
+    dataset.push_to_hub(unfiltered_dataset_name)
+    
+    def filter_func(example):
+        rewards = example["filter_rewards"]
+        # get the mean of the rewards that are not None
+        mean_reward = torch.nanmean(torch.tensor(rewards, dtype=torch.float32))
```
lewtun, line 164: This can be done above when we include a `pass_rate` column and then this filter could be a lambda function

scripts/filter_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+    dataset = dataset.map(batch_score, batched=True, batch_size=256)
+    
+    
+    if script_args.output_dataset_name is not None:
+        unfiltered_dataset_name = script_args.output_dataset_name
+    else:
+        unfiltered_dataset_name = f"{script_args.dataset_name}-generated"
+    filtered_dataset_name = f"{unfiltered_dataset_name}-filtered"
+    
+    dataset.push_to_hub(unfiltered_dataset_name)
```
lewtun, line 159: Instead of having a proliferation of datasets, how about we use dataset configs to denote the unfiltered vs filtered subsets, e.g.:

* `unfiltered` = this one
* `filtered_{script_args.low_reward_threshold}-{script_args.high_reward_threshold}`

That way we can do future filterings and have them all in one place.

scripts/filter_dataset.py
```diff
@@ -0,0 +1,179 @@
...
+            
+        return examples
+        
+    dataset = dataset.map(batch_score, batched=True, batch_size=256)
+    
+    
+    if script_args.output_dataset_name is not None:
+        unfiltered_dataset_name = script_args.output_dataset_name
+    else:
+        unfiltered_dataset_name = f"{script_args.dataset_name}-generated"
```
lewtun, line 156:
```suggestion
        unfiltered_dataset_name = f"{script_args.dataset_name}-pass-rate"
```

recipes/dataset_filtering/config_demo.yaml
```diff
@@ -0,0 +1,27 @@
...
+temperature: 0.7
+
+# Reward func arguments
+reward_funcs:
+- accuracy
+reward_weights:
+- 1.0
+
+# Filtering arguments. Samples with mean reward outside of low / high will be filtered
+low_reward_threshold: 0.2
```
lewtun, line 26: Technically we filter on pass rate, so let's use that here:
```suggestion
pass_rate_min: 0.2
```

recipes/dataset_filtering/config_demo.yaml
```diff
@@ -0,0 +1,27 @@
...
+high_reward_threshold: 0.8
```
lewtun, line 27:
```suggestion
pass_rate_max: 0.8
```

recipes/dataset_filtering/config_demo.yaml
```diff
@@ -0,0 +1,27 @@
...
+num_generations: 4
+temperature: 0.7
+
+# Reward func arguments
+reward_funcs:
+- accuracy
+reward_weights:
+- 1.0
+
+# Filtering arguments. Samples with mean reward outside of low / high will be filtered
```
lewtun, line 25:
```suggestion
# Filtering arguments. Samples with a pass rate outside the interval `pass_rate_min < x < pass_rate_max` will be filtered.
```

lewtun (approved):
LGTM with some small suggestions to combine the subsets as configs and align the variables with pass rate

scripts/filter_dataset.py
edbeeching: I was thinking to change the dataset name to include the model and revision instead?

scripts/filter_dataset.py
edbeeching: I will expose it, but I think we should be using the same generation parameters here as we used for RL, rather than evaluation? As that is what we will use the samples for.

PR #639, opened 2025-05-09
Add time to Slurm

lewtun:
Add `--time` flag to comply with new HF Cluster requirements: https://huggingface.slack.com/archives/C04LMQCANLS/p1746797087651049

I only edited the Slurm files that did not have a preexisting `--time` flag set.

### DIFF
--- a/slurm/e2b_router.slurm
+++ b/slurm/e2b_router.slurm
@@ -6,6 +6,7 @@
 #SBATCH --output=/fsx/open-r1/logs/e2b_router/%x-%j.out
 #SBATCH --error=/fsx/open-r1/logs/e2b_router/%x-%j.err
 #SBATCH --requeue
+#SBATCH --time=7-00:00:00
 
 echo "Starting job"
 set -x -e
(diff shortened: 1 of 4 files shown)


slurm/e2b_router.slurm
```diff
@@ -6,6 +6,7 @@
 #SBATCH --output=/fsx/open-r1/logs/e2b_router/%x-%j.out
 #SBATCH --error=/fsx/open-r1/logs/e2b_router/%x-%j.err
 #SBATCH --requeue
+#SBATCH --time=7-00:00:00
```
lewtun, line 9: I took a best guess here, but feel free to reduce / increase as you see fit

slurm/evaluate.slurm
```diff
@@ -5,6 +5,8 @@
 #SBATCH --output=./logs/%x-%j.out
 #SBATCH --error=./logs/%x-%j.err
 #SBATCH --requeue
+#SBATCH --time=1-00:00:00
```
lewtun, line 8: I don't think we have any evals that run longer than 1 day, but let me know if we should increase it

edbeeching (approved):
Thanks, the times seem reasonable. LGTM

PR #662, opened 2025-05-27
Refresh Weka on Slurm

lewtun:
Only targets the h4 cache for now.


edbeeching (commented):
Do we not want to fetch the open-r1 dir? Perhaps we can exclude the data/ directory.

PR #663, opened 2025-05-27
Align EOS token ID between tokenizer and generation config

lewtun:
This PR addresses a nasty footgun in `transformers`, where a change to the base model's tokenizer must ALSO be propagated to the model's generation config. Without this change, the `pipeline()` function produces unbounded generations because it relies on the base model's EOS token (e.g. `<|endoftext|>`) instead of the one we set in the tokenizer config (e.g. `<|im_end|>`)

Note that this has no impact on `vllm` since there the EOS token is inferred from the `tokenizer_config.json` file.

cc @qgallouedec @kashif for viz as this might be necessary to include on the SFT script of TRL too.

I also took the opportunity to clean up some of the "demo" recipes as I think it's better to have a single source of truth for well-tested recipes.


src/open_r1/grpo.py
```diff
@@ -140,6 +140,9 @@ def make_conversation(example, prompt_column: str = script_args.dataset_prompt_c
     # Save model and create model card
     ##################################
     logger.info("*** Save model ***")
+    # Align the model's generation config with the tokenizer's eos token
+    # to avoid unbounded generation in the transformers `pipeline()` function
+    trainer.model.generation_config.eos_token_id = tokenizer.eos_token_id
```
lewtun, line 145: Here is the key change to avoid the generation footgun

kashif:
ok checking

edbeeching (approved):

